PR #1, opened 2024-03-25
Bt15 green

Michael8818:
Please review and lets start discussing any other bugs or issues we find.

@burkewm @albertvasconcellos


burkewm (approved):
Rosemon X issue is resolved, effects of all Green Digimon, Eggs, Tamers and Options are all working as intended.

PR #2, opened 2024-03-30
Bt15 black

albertvasconcellos:
Please review and let me know if there are any bugs that need to be fixed

@burkewm @Michael8818


Michael8818 (changes requested):
First round of testing I have found the following:

Ryudamon
- inherited doesn't trigger

Ginryumon - 
- inherited works for any of your digimon to suspend, 
- not once per turn.

Hisyaryumon
- inherited not working

Warumonzaemon 
- throws error when no target available

Ouryumon 
- can suspend any permanent (not just digimon/tamers)

Ultimate slicer 
- does not work

Assets/CSV/.~lock.BT15.csv#
Michael8818, line 1: Lets remove this lock file. As a precaution

Michael8818 (changes requested):
Have not retested but the code looks good to fix the issue.

Also for other colors be careful not to make ANY changes to the openings scene. Mine comes up with changes all the time, I just discard them. we don't want to deal with merge conflicts due to the unity scene.

PR #3, opened 2024-04-02
Bt15 purple

Michael8818:
@albertvasconcellos @burkewm


burkewm (approved):
All purple cards tested and validated. Levia X tested multiple times to ensure it's effects were activating as needed. Myotismon line and others working as expected.

PR #6, opened 2024-04-09
Bt15 Yellow

albertvasconcellos:
@Michael8818 @burkewm


albertvasconcellos:
1. The second part of Celestial Arrow's effect doesn't ask if you want to trash a security. (Not optional)

2. Babamon's end of opponent's turn effect not prompting. (Not optional)

PR #7, opened 2024-04-16
St17 PR

albertvasconcellos:
@Michael8818 @burkewm


Assets/CardEffect/ST17/Green/Cherubimon_ST17_09.cs
```diff
@@ -0,0 +1,281 @@
...
+                                List<SelectionElement<bool>> selectionElements = new List<SelectionElement<bool>>()
+                        {
+                            new SelectionElement<bool>(message: $"From hand", value : true, spriteIndex: 0),
+                            new SelectionElement<bool>(message: $"From trash", value : false, spriteIndex: 1),
+                        };
+
+                                string selectPlayerMessage = "From which area do you play a card?";
+                                string notSelectPlayerMessage = "The opponent is choosing from which area to play a card.";
+
+                                GManager.instance.userSelectionManager.SetBoolSelection(selectionElements: selectionElements, selectPlayer: card.Owner, selectPlayerMessage: selectPlayerMessage, notSelectPlayerMessage: notSelectPlayerMessage);
```
Michael8818: Have not tested this but wouldn't this mean you won't be able to select something to delete unless you have a target to play in trash, when realistically they are both may effects so neither depend on each other to trigger or activate.

Michael8818 (changes requested):
Overall Code looks good, I would make note of some method names could be more clear. Having UseCondition is fine, but having UseCondition1, UseCondition2, etc.... could be named better.

Michael8818:
Terriermon - can not trigger main effect
Wendigomon - deletion triggered, but did not play 1 from trash
Megagargomon ace - triggered at end of any attack
Rapidmon ST17-06 - can select any permanent to receive dp -
Giant Missile - does not prevent unsuspend
Henry Wong - Can trigger its main effect

albertvasconcellos:
Megagargo Ace can't digivolve over level 4 ST17 Rapidmon but it can digivolve over BT8 Rapidmon (Not seeing Rapidmon name)
Megagargo Ace given option to unsuspend after a permanent digivolves or is played (Timing issue)
Cant digivolve or unsuspend tooltip doesnt stay when turn ends

Rapidmon ST17 security reduction is only 2k not 4k
Fire Rocket not recognizing ST17 Rapidmon as Armor Form

Double typhoon promting to add cards even though it isn't optional
double typhoon not going to battle area after security check (effect works though)

Gargomon effect triggering when anything suspends

cherubimon when digivolve didnt trigger second effect when I had no target to delete

Fire rocket not recognizing Armor form in ST17 magnamon

Movie Veemon not reducing cost of digivolving into ST17 magnamon
ST17 Magna not triggering second part of when digivolving unless the first part happens

albertvasconcellos:
Megagargomon Ace stops egg from digivolving

PR #46, opened 2024-10-17
Bt18 purple

Erdandrej:
Implemented all the remaining purple card effects. 
LordKnightmon Ace is still missing the static collision effect; it might be better to implement it outside of this branch because it requires changing the collision classes and then all other cards that have collision.


Assets/CardEffect/BT18/Purple/Duskmon_BT18_078.cs
```diff
@@ -0,0 +1,461 @@
...
+
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.IsTamer && targetPermanent.TopCard.CardColors.Contains(CardColor.Purple);
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(
+                    permanentCondition: PermanentCondition, digivolutionCost: 3, ignoreDigivolutionRequirement: false,
+                    card: card, condition: Condition));
+            }
```
Michael8818, line 31: What is this for? it should just be a standard digivolution attached to the scriptable object

Michael8818 (approved):

PR #51, opened 2024-11-20
LM Memory Boosts and Scrambles

Quick-Bolt:


Michael8818 (changes requested):
LGTM, though I noticed the scramble cards delay effect doesn't destroy the option card. Simple copy from the other boost cards will suffice.

Michael8818 (approved):

PR #55, opened 2025-01-08
EX8 and related promos

jdelia1:


Assets/CardEffect/P/Black/RookDevice_P_159.cs
Michael8818, line 1: Why was these changes made?

jdelia1: the block I modified was the trashed from battle area block, but it had a bunch of references to main effect, and it would try to place the card in the battle area at the end of its execution.

PR #66, opened 2025-01-29
2.5 Purple Ghosts

Erdandrej:


Assets/Scripts/CardEffectCommons/KeyWordEffects/Overclock.cs
Michael8818, line 1: why was overclock changed to end of turn timing?

Erdandrej: I realized the give methods for the keywords I made in the past had wrong timings so I updated them; overclock should be end of turn like vortex based on its wording. No cards used those methods, but I was future-proofing since I noticed it while making execute.

PR #74, opened 2025-02-10
Bt19beelzgargofixes

dp101428:
Fixed traits for gargomon (missing beastkin, had digimon instead), beelzemon (misssing 7gdl), and fixed beelz's effect duration (was set to UntilEachTurnEnd instead of UntilOpponentTurnEnd)


dp101428:
tried to merge to wrong branch, please ignore

PR #82, opened 2025-03-14
All ST20 cards

dp101428:
All cards should be fully implemented. Apologies in advance for any errors


Assets/CardEffect/ST20/Black/ST20_10.cs
```diff
@@ -0,0 +1,88 @@
...
+            #endregion
+
+            #region Warp Digivolution
+            if (timing == EffectTiming.None)
+            {
+                bool enoughTamerColours()
+                {
+                    List<CardSource> tamerCards = new List<CardSource>();
+
+                    foreach (Permanent permanent in card.Owner.GetBattleAreaPermanents())
```
Michael8818, line 32: Use CardEffectCommons.PermanentMatchConditionCount instead here. I think that is the name of method but could be slight off

Assets/CardEffect/ST20/Black/ST20_10.cs
```diff
@@ -0,0 +1,88 @@
...
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent == card.PermanentOfThisCard();
+                }
+
+                bool CardCondition(CardSource cardSource)
+                {
+                    if (cardSource != null)
+                    {
+                        //compared to the code I was copying from, removed a check that owner matched between cardsource and card, as that should be handled by whose hand the card is in. If this breaks, try putting it back.
```
Michael8818: If you can add what card you were copying from

Assets/CardEffect/ST20/Black/ST20_10.cs
```diff
@@ -0,0 +1,88 @@
...
+                        if (card.Owner.HandCards.Contains(cardSource))
+                        {
+                            return cardSource.EqualsCardName("WarGreymon");
+                        }
+                    }
+
+                    return false;
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 4, ignoreDigivolutionRequirement: true, card: card, condition: Condition, cardCondition: CardCondition));
```
Michael8818, line 74: This card effect factory here only lets you digivolve into the target previously mentioned. It doesn't actually digivolve is that the intent

Assets/CardEffect/ST20/Black/ST20_11.cs
```diff
@@ -0,0 +1,397 @@
...
+
+                foreach (Permanent permanent in card.Owner.GetBattleAreaPermanents())
+                {
+                    if (permanent.IsTamer)
+                    {
+                        tamerCards.Add(permanent.TopCard);
+                    }
+                }
+                return Combinations.GetDifferenetColorCardCount(tamerCards) / 2;
+            }
```
Michael8818, line 56: this could be replaced with CardEffectCommons.MatchConditionPermanentCount

Assets/CardEffect/ST20/Green/ST20_07.cs
```diff
@@ -0,0 +1,85 @@
...
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+            #region alternative digivolution method
+            if(timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent permanent)
+                {
+                    return permanent.Level == 2 && permanent.TopCard.EqualsTraits("ADVENTURE");
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(PermanentCondition, 2, false, card, null));
```
Michael8818: should be for 0 cost

Assets/CardEffect/ST20/Green/ST20_08.cs
```diff
@@ -0,0 +1,108 @@
...
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 2, ignoreDigivolutionRequirement: false, card: card, condition: null));
+            }
+            #endregion
+
+            #region When Digivolving
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Play 1 Tamer with the [ADVENTURE] trait name from hand", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, true, EffectDescription());
```
Michael8818: go ahead and make this mandatory

Assets/CardEffect/ST20/Green/ST20_09.cs
```diff
@@ -0,0 +1,305 @@
...
+                List<CardSource> tamerCards = new List<CardSource>();
+
+                foreach (Permanent permanent in card.Owner.GetBattleAreaPermanents())
+                {
+                    if (permanent.IsTamer && permanent.TopCard.EqualsTraits("ADVENTURE"))
+                    {
+                        tamerCards.Add(permanent.TopCard);
+                    }
+                }
+                return Combinations.GetDifferenetColorCardCount(tamerCards) / 2;
```
Michael8818, line 44: can use CardEffectCommons.MatchConditionPermanentCount

Assets/CardEffect/ST20/Red/ST20_04.cs
```diff
@@ -0,0 +1,325 @@
...
+            {
+                List<CardSource> tamerCards = new List<CardSource>();
+
+                foreach (Permanent permanent in card.Owner.GetBattleAreaPermanents())
+                {
+                    if (permanent.IsTamer)
+                    {
+                        tamerCards.Add(permanent.TopCard);
+                    }
+                }
```
Michael8818, line 42: can use CardEffectCommons.MatchConditionPermanentCount instead

Michael8818 (changes requested):

Assets/CardEffect/ST20/Black/ST20_10.cs
dp101428: Not entirely sure how the proposed alternative is meant to work. The count here is the number of distinct colours between all the tamers, if I just counted the number of tamers it wouldn't work.

Assets/CardEffect/ST20/Black/ST20_10.cs
dp101428: Couldn't find it unfortunately (I thought it was bt14 wargreymon but it has no such line of text). Could remove comment if that's preferred to being unable to clarify which card. 

Edit: Nvm, found it, ST8-04 Veemon (the warp one), figure the code is probably just weird because it's such an old card (and there aren't many non-triggered warp digivolution effects where the effect to digivolve is actually on the rookie)

Assets/CardEffect/ST20/Black/ST20_10.cs
dp101428: The intent is to not actually digivolve yes. This is a static effect that it can just do, like BT2 Beelz, it's not an activated/triggered effect like BT17 Dorumon

Assets/CardEffect/ST20/Black/ST20_11.cs
dp101428: Already addressed regarding the way ST20-10 Agumon counts

Assets/CardEffect/ST20/Green/ST20_07.cs
dp101428: Good catch, thanks

Assets/CardEffect/ST20/Green/ST20_08.cs
dp101428: Not sure why it needs to be, I guess because currently the player gets 2 chances to back out, both when choosing to trigger and choosing what to play? Fair enough, changing

PR #86, opened 2025-03-19
2.5 Bug fixes

jdelia1:
King Drasil, Dorugoramon, Taomon, Ryugumon, Ryo Akiyama, Pinamon, Gankoomon, Singularity of Chaos

### DIFF
--- a/Assets/CardEffect/BT16/Yellow/BT16_102.cs
+++ b/Assets/CardEffect/BT16/Yellow/BT16_102.cs
@@ -90,8 +90,7 @@ namespace DCGO.CardEffects.BT16
                     if (selectedPermanent != null)
                     {
                         if (card.PermanentOfThisCard().DigivolutionCards.Count((cardSource) =>
-                                cardSource.ContainsCardName("Magnamon (X Antibody)") ||
-                                cardSource.ContainsCardName("Magnamon(X Antibody)") ||
+                                cardSource.EqualsCardName("Magnamon (X Antibody)") ||
                                 cardSource.ContainsTraits("Armor Form")) >= 1)
                         {
                             CanNotAffectedClass canNotAffectedClass = new CanNotAffectedClass();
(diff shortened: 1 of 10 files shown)


Assets/CardEffect/BT16/Yellow/BT16_102.cs
```diff
@@ -89,7 +89,10 @@ IEnumerator ActivateCoroutine(Hashtable _hashtable)
 
                     if (selectedPermanent != null)
                     {
-                        if (card.PermanentOfThisCard().DigivolutionCards.Count((cardSource) => cardSource.ContainsCardName("Magnamon (X Antibody)") || cardSource.ContainsTraits("Armor Form")) >= 1)
+                        if (card.PermanentOfThisCard().DigivolutionCards.Count((cardSource) =>
+                                cardSource.ContainsCardName("Magnamon (X Antibody)") ||
+                                cardSource.ContainsCardName("Magnamon(X Antibody)") ||
```
Michael8818: don't need both of these, just the one with spaces. also use EqualsCardName instead. Also if fixing the name of a card is the real issue, do so on the actual cards entity instead.

Assets/CardEffect/BT19/Blue/BT19_027.cs
```diff
@@ -177,7 +177,8 @@ IEnumerator SelectCardCoroutine(Permanent permanent)
 
                     if (selectedPermanent != null)
                     {
-                        int selectedLevel = selectedPermanent.Level;
+                        // If selected digimon has no level, set reference level to 0.
+                        int selectedLevel = selectedPermanent.TopCard.HasLevel ? selectedPermanent.Level : 0;
```
Michael8818: instead of this just change the line 178 if to be:

if(selectedPermanent != null && selectedPermanent.TopCard.HasLevel)

because we really don't want to run this code if the selected card doesn't have a level at all

Michael8818 (changes requested):

Michael8818 (approved):

PR #87, opened 2025-03-20
1.7.6 Bug Fixes

LewisWelch94:
### Cards
BT2-024 Seadramon - Alt art Trait.
BT2-066 Machinedramon - Using modified Degeneration class (changes listed below)
BT11-074 BlackWarGreymon (X Antibody) - Using modified isMinCost (Listed below)
BT17-093 Tai Kamiya & Kari Kamiya - fixed typo
BT17-098 Hacker Pride - Added ACE check, and triggers overflow inside adding to security step.
BT18-042 MagnaGarurumon - Added ACE check, and triggers overflow inside adding to security step.
BT18-052 CannonBeemon - now gets Degenerate count from Flipped up Sec cards & using modified Degeneration class (changes listed below)
BT18-072 AncientBeetlemon - Using modified Degeneration class (changes listed below)
BT19-011 WarGrowlmon ACE - Added check to see which of the selected digimon still remains on field, and use that to get correct memory gain count.
BT19-090 Meteor Rock Soul - Edited CanTarget checks, now using CanUnsuspend + doing currently digimon on field check, so if there is 1 or less of the required, effect ends.
EX4-069 Gaia Reactor - Using modified isMaxCost (Listed below)
EX8-029 Aegisdramon - Fixed effect description
EX8-033 Pumpkinmon - Added check to see if its an digiegg, if true use AddLibraryBottomCards, else normal add to hand.
EX8-052 Cyberdramon (X Antibody) - Edited canActivate to only check for option, added opponent digimon check in the ActivateCoroutine.
P-042 Gabumon - Fixed typo and incorrect Card Effect Class name in alt arts.
ST12-03 Solarmon - Fixed incorrect Card Effect Class name in alt art.
EX7-008 ToyAgumon - Fixed typo in Card Selection title

### Methods
IDegeneration - Added optional boolean paramater, which is defaulted to null, used when we want to De-Digivolve count to happen before the target selection as per some card rulings.
IsMaxCount & IsMinCost - Cleaned up the filtering & select & place the return inside the if/else statement.
having the return inside, is mainly used for IsDigimonOnly, as return outside means it loses this requirement, allowing tamers to be selectable if they are equal to/are the min/max cost card.


Assets/CardEffect/EX8/Yellow/EX8_033.cs
Michael8818, line 1: this isn't the right way to fix this. it should involve fixing the SelectCardEffect class. i would rather just scrap this and fix the select class instead

Assets/CardEffect/P/Blue/P_042.cs
```diff
@@ -0,0 +1,70 @@
+using System.Collections;
+using System.Collections.Generic;
+
+namespace DCGO.CardEffects
```
Michael8818: missing P in namespace

Assets/CardEffect/P/Blue/P_042.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public class P_042 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            if (timing == EffectTiming.None)
```
Michael8818: incorrect timing

Assets/CardEffect/P/Blue/P_042.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Reveal top 5 cards of deck", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, true, EffectDiscription());
```
Michael8818: effect is not optional

Assets/Scripts/CardController.cs
Michael8818, line 1: correct me if I am wrong here but having DegenerationCountRuling not being null means you won't select the count to dedigivolve? but in reality what the issue with the cards mentioned is you shouldn't be able to select dedigivolve amount per card (in the case of multiple cards dedigivolved). This looks like it would bypass the the selection of the count entirely? am I missing something?

Michael8818 (changes requested):

Assets/Scripts/CardController.cs
LewisWelch94: Unless im misunderstanding, the issue was the ruling stated you had to select your de-digivolve count first, then select your targets & De-Digivolve both of them by the same amount pre-selected, hence we add in the optional boolean, so it bypasses the de-digivolve count inside IDegeration, and uses the one we pass in instead for all targets

From bug report
> Currently is allowing you to individually choose how much to de-digivolve from each digimon. Newly released rulings say you must choose how many to de-digivolve and then apply that same amount to both cards.

Michael8818 (approved):

PR #104, opened 2025-04-06
Bt21 Red -  029 + Token

LewisWelch94:


Assets/CardEffect/BT21/Red/BT21_029.cs
```diff
@@ -0,0 +1,249 @@
...
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    if (CardEffectCommons.IsExistOnBattleAreaDigimon(card))
+                    {
+                        if (CardEffectCommons.CanTriggerOnPermanentDeleted(hashtable, SharedPermanentCondition))
+                        {
+                            return true;
+                        }
```
Michael8818: this should be checking if a security stack is removed not if a permanent was deleted

Michael8818 (changes requested):

Michael8818 (approved):

PR #132, opened 2025-05-13
Ex9 002

Michael8818:


Michael8818:
Made a mistake here and accidentally uploaded EX9-002 instead of EX9-001.

Assets/CardEffect/EX9/Red/EX9_001.cs
```diff
@@ -0,0 +1,85 @@
...
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+                }
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    Permanent selectedPermanent = card.PermanentOfThisCard();
+
+                    if (selectedPermanent != null && selectedPermanent.HasFaceDownDigivolutionCards)
```
LewisWelch94: Seems more in the ruling that `HasFaceDownDigivolutionCards` is in the CanActivateCondition.
as while they both have the same outcome when there is a FD source, it would differ when there isnt one

LewisWelch94 (approved):

PR #155, opened 2025-06-09
P188189

dp101428:
implemented P-188 Demiveemon and P-189 Dimetromon

### DIFF
--- a/Assets/CardEffect/P/Red/P_189.cs
+++ b/Assets/CardEffect/P/Red/P_189.cs
@@ -1,0 +1,216 @@
+using System.Collections;
+using System.Collections.Generic;
+
+//Dimetromon
+namespace DCGO.CardEffects.P
+{
+    public class P_189 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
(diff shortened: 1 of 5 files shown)


Assets/CardEffect/P/Red/P_189.cs
```diff
@@ -0,0 +1,216 @@
...
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnExecutingArea(card);
+                }
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                {
+                    return (cardSource.ContainsTraits("Liberator") || cardSource.ContainsTraits("LIBERATOR")) &&
```
Michael8818, line 41: nothing wrong here but FYI, ContainsTraits will compare traits as lower case and remove unnecessary characters. so you can just use the trait that is on the card itself no need for multiple versions.

Michael8818 (approved):

Assets/CardEffect/P/Red/P_189.cs
dp101428: Ah yeah this was just a consequence of copy/pasting from older card and not checking properly, thanks for the reminder.

PR #156, opened 2025-06-09
190, 191, 192, 193

LewisWelch94:


Assets/CardEffect/P/Purple/P_192.cs
Michael8818, line 1: add card name as comment above namespace (see P_190)

Michael8818 (changes requested):

Michael8818 (approved):

PR #159, opened 2025-06-10
Bug Fixes

LewisWelch94:


Assets/CardEffect/EX9/Blue/EX9_019.cs
```diff
@@ -280,7 +286,7 @@ bool PlayedRequirement(Permanent permanent)
                 {
                     return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) &&
                            !permanent.IsOption &&
-                           (permanent.TopCard.HasGreymonName || permanent.TopCard.ContainsCardName("Matt Ishida"));
+                           (permanent.TopCard.HasGreymonName || permanent.TopCard.CardNames.Some((cardName) => cardName.Contains("Matt Ishida") || cardName.Contains("MattIshida")));
```
Michael8818, line 289: we should look at this further, because ContainsCardName does the exact same thing you are doing here. Would much rather prefer to use the method created instead. must be something off about the card data itself here

Assets/CardEffect/EX9/Red/EX9_012.cs
```diff
@@ -191,7 +195,7 @@ bool PlayedRequirement(Permanent permanent)
                 {
                     return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) &&
                            !permanent.IsOption &&
-                           (permanent.TopCard.HasGarurumonName || permanent.TopCard.ContainsCardName("Tai Kamiya"));
+                           (permanent.TopCard.HasGarurumonName || permanent.TopCard.CardNames.Some((cardName) => cardName.Contains("Tai Kamiya") || cardName.Contains("TaiKamiya")));
```
Michael8818, line 198: Same as EX9-019

Assets/CardEffect/EX9/Red/EX9_066.cs
```diff
@@ -70,21 +71,23 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
-                    IEnumerator SelectCardCoroutine(CardSource source)
+                    IEnumerator AfterSelectCardCoroutine(List<CardSource> sources)
                     {
-                        selectedSource = source;
+                        selectedSource = sources[0];
                         yield return null;
                     }
 
-                    if(selectedSource == null)
+                    if (selectedSource == null)
```
Michael8818: this is the exact same as what it was previously just slightly different timing.

Assets/CardEffect/EX9/Yellow/EX9_025.cs
Michael8818, line 1: remove line 116/117 (selectedCard.SetReverse();) add isFaceDown: true to AddDigivolutionCardsBottom call below this

Michael8818 (changes requested):

Assets/CardEffect/EX9/Blue/EX9_019.cs
LewisWelch94: Hmm i see, I dont have access to card data. may need a chat with you about this then.

Assets/CardEffect/EX9/Red/EX9_012.cs
LewisWelch94: Same as above reply

Michael8818 (approved):

PR #191, opened 2025-08-23
Adding once-per-turn check to CanTrigger (BT22-052 & EX9-056)

skikklesman:
Adding the same OPT check that CanActivate has to CanTrigger.

Context: The autoProcessor_CutIn processor doesn't update as much as the regular autoProcessor, which regularly cleans its queue of stacked skills. In any scenario where autoProcessor_CutIn .StackSkillInfos adds triggered effects to its list of stacked effects, but those effects can't activate right away (like if the effect is OPT and has already been used), the triggered effect stays in the list and gets stale, until it is cleaned out at the next action.  If that next action happens on the opponent's turn, the OPT check is cleared and the effect can activate.  Adding the check in CanTrigger stops that effect from ever getting added if it's already been used the max number of times that turn.

### DIFF
--- a/Assets/Scripts/ICardEffect.cs
+++ b/Assets/Scripts/ICardEffect.cs
@@ -288,6 +288,15 @@ public abstract class ICardEffect
 
         #endregion
 
+        #region Effect availability determined by the maximum number of times it can be used in a turn
+
+        if (EffectSourceCard.cEntity_EffectController.isOverMaxCountPerTurn(this, MaxCountPerTurn))
+        {
+            return false;
+        }
+
+        #endregion
+
         #region Determination of availability for each effect
 
         if (CanUseCondition == null || !CanUseCondition(hashtable))


Michael8818 (approved):

Michael8818:
> Adding the same OPT check that CanActivate has to CanTrigger.
> 
> Context: The autoProcessor_CutIn processor doesn't update as much as the regular autoProcessor, which regularly cleans its queue of stacked skills. In any scenario where autoProcessor_CutIn .StackSkillInfos adds triggered effects to its list of stacked effects, but those effects can't activate right away (like if the effect is OPT and has already been used), the triggered effect stays in the list and gets stale, until it is cleaned out at the next action. If that next action happens on the opponent's turn, the OPT check is cleared and the effect can activate. Adding the check in CanTrigger stops that effect from ever getting added if it's already been used the max number of times that turn.

Funny the way this reads sounds like a lot of other issues we have had about effects triggering on the next action after they should (attacking by effect for example). Seems like you may have also found an issue there that could be corrected. I will merge this but I want to keep this PR open so we can reference later.

PR #199, opened 2025-09-08
Appfuse simplification

dp101428:
These functions serve to reduce a lot of the repetitive code that needs to be written to ensure that each appfusion effect makes the cards being checked meet different requirements from the same list, which is a part of the mechanic that isn't card-specific and so seems useful to codify. It's in CardEffectFactory because it seemed the most similar to AddSelfDigivolutionRequirement which is also in that class. To use it to implement any of the current appfusions, you can just pass it a list of strings for the card names, and of course the card variable itself.

So for example, to add Mienumon's appfuse condition, you'd simply do `cardEffects.Add(CardEffectFactory.AddAppfuseMethodByName(new List<string>(){"Mirrormon", "Kabemon", "Copipemon"}, card))`

Hopefully this'll cause fewer card-by-card issues with appfusion since there'll be fewer places to make a mistake.


Michael8818 (commented):
This looks alright to me but before it is merged I would like it to be tested on a card.

dp101428:
Test has been completed, glad I did it because I'd have missed a bug otherwise, but it all works now.

PR #608, opened 2026-02-14
Update BT22_064.cs

x89rt2:
AT only triggering if oppo has digimon. Should always trigger.


Assets/Scripts/CardEffect/BT22/Black/BT22_064.cs
```diff
@@ -139,8 +139,7 @@ bool CanUseCondition(Hashtable hashtable)
 
                 bool CanActivateCondition(Hashtable hashtable)
                 {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
```
hooperk, line 142: I can't check on mobile, is an extra check needed in the coroutine if we aren't checking here is is it already safe?

LewisWelch94 (approved):

PR #609, opened 2026-02-15
Update EX11_054.cs

x89rt2:
Fix AT not giving DP if 2 or more valid targets.


x89rt2:
I can't tell the difference between the 2 selectedPermanent but the IDE did so fixed it.

LewisWelch94 (approved):

PR #620, opened 2026-02-17
CEF methods for creating triggers for effects

hooperk:
While there are a couple of cards making use of some of this, the biggest proof of these methods will be actually using them in developing new cards. To do that we need this in before we start working on new cards to actually make use of it.


x89rt2:
I have merged the cardeffectsfactory only. This can be closed.

PR #622, opened 2026-02-18
Removed RPCs from Select Command Effects

BennAbbot:
Updated some old effects, that where using there own RPCs, to use the `UserSelectionManager`


hooperk (approved):

hooperk (approved):
LGTM

PR #624, opened 2026-02-18
Implement ST24-07 and proposal for how dual cards can handle

hooperk:
TODO:

- [x] Arts Digivolution is not by effect. Need to separate it from activateClass.
- [ ] Change all instances of CardSource.CardColors.Contains to CardSource.HasCardColor everywhere. Not going to this effort until we confirm the approach (I will do this change, just need confirmation first so I am not wasting time)
- [x] Make it so when player drags a DualCard that is seen as using the option
- [x] Update json parsing to populate DualCardColors when present. and to recognise when cardkind is dualcard
- [x] Tweaking to the right click window and deck building screen to display all information.
- [ ] Further changes to cEntity_Base to store the separate effect text for the option half.
- [ ] Update cards which trash Delay options such as BT19-064 Justimon, BT18-100 Gospel of the Fallen Angel to NOT delete DualCards, depending on rulings
- [x] Rework for CardKind being a list

I am not suggesting all of the above will be done by me, just trying to track all things likely to be needed

### DIFF
--- a/Assets/Scripts/Script/CardController.cs
+++ b/Assets/Scripts/Script/CardController.cs
@@ -301,6 +301,7 @@ public class PlayCardClass
     {
         bool burstDigivolved = false;
         bool appFusion = false;
+        bool isEvolution = false;
 
         List<CardSource> playedCards_fixed = new List<CardSource>();
 
@@ -388,8 +389,6 @@ public class PlayCardClass
             #endregion
 
             #region Determine if Evolution
-
-            bool isEvolution = false;
 
             if (targetPermanents.Count >= 1)
             {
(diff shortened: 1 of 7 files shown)


Assets/Scripts/Script/CardController.cs
```diff
@@ -1752,6 +1752,11 @@ public IEnumerator UseOption()
 
             #endregion
 
+            if (card.Owner.ExecutingCards.Contains(card) && card.IsDualCard)
+            {
+                yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.ArtsDigivolve(card));
+            }
```
hooperk, line 1758: Depending on rulings this may have to move to lower priority, by being here it becomes a choice to digivolve first and if you don't to then add to security if using via Sakuyamon

x89rt2:
Is this still relevant with your other pull?

hooperk:
> Is this still relevant with your other pull?

Only for the checklist tbh

hooperk:
Completed elsewhere

PR #627, opened 2026-02-18
General fix: Multiple DigiXros or Assemblies at once

hooperk:
Basic idea is rather than each of these classes representing exactly 1 digixros or assembly, they maintain a mapping between a particular card and a digixros or assembly attempt for that card, since the last time the class was reset.

The classes now also maintain a history of all cardSources already selected to be placed under cards by either class so they do not get selected again and cause conflict.

This should enable multiple digixrosses when several cards are all played at once (BT5-087 Omnimon Zwart and 2 digixros capable black and purple can test, and EX9-047 Eyesmon can test Assembly while being played alongside something else, or even 2 Eyesmon with 8 Shatter mode in trash for full test)

Has been tested to ensure it does not break standard digixros and assembly. Will need to be tested with playing 2+ cards at once of which 1 or more has assembly or digixros, and a double check over the 4 cards that needed to be changed.

### DIFF
--- a/Assets/Scripts/CardEffect/BT10/Purple/BT10_093.cs
+++ b/Assets/Scripts/CardEffect/BT10/Purple/BT10_093.cs
@@ -114,7 +114,7 @@ namespace DCGO.CardEffects.BT10
             {
                 if (CardEffectCommons.IsExistOnBattleArea(card))
                 {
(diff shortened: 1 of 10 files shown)


x89rt2:
So why does this one change the following cards?

1 Yuu Amano                    BT10-093 
1 Shoutmon X7: Superior Mode   BT12-112 
1 Apocalymon                   BT15-102 
1 Shoutmon X7: Superior Mode   BT21-030 
1 Apocalymon                   EX10-061

hooperk:
Those cards slotted a card to be placed upon the next digixros. They needed to change to slotting a card to be placed upon the next digixros for the correct card

x89rt2 (approved):

PR #629, opened 2026-02-18
Reset used blitz if a new open state is reached in main phase

### LINKED ISSUE #478: Complicated Blitz maneuvers are not working fully correctly
Very specific Gallantmon combos fails to get the second Blitz
[Video](https://discord.com/channels/1095717982522585098/1229224709007343687/1279440121963483200)

Maybe also this Jesmon GX Seiken Meppa play?:
[Bug Report](https://discord.com/channels/839041812932329512/1189361807727542393/1287373033022951475)

Blitz, end of turn warp to rush gallant with EX2 takato giving it blitz to try and blitz again. Video sort of looks like the end of turn is activating mid attack so it doesn't let you declare the attack?
[Bug Report](https://discord.com/channels/839041812932329512/1288213415877410968)

hooperk:
Blitz couldn't be activated twice in one turn even in valid scenarios

### DIFF
--- a/Assets/Scripts/Script/TurnStateMachine.cs
+++ b/Assets/Scripts/Script/TurnStateMachine.cs
@@ -920,6 +920,14 @@ public class TurnStateMachine : MonoBehaviourPunCallbacks
             ResetMainPhaseParameter();
             #endregion
 
+            #region resetting used blitz
+            if (GManager.instance.attackProcess.UsedBlitz)
+            {
+                if (GManager.instance.turnStateMachine.gameContext.Memory <= 0)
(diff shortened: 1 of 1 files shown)


x89rt2:
I assume there's no interaction between this and the attack changes?

hooperk:
> I assume there's no interaction between this and the attack changes?

Correct, this happens at a later point in the entire main phase process where the attack should have completed. TBH I'm not sure the bool for blitz being used is needed anymore with other changes we've made to prevent 2 EOT attacks at once, but this is a simple fix for resetting it and leaving it in that I can be more confident will continue to work correctly

x89rt2 (approved):

PR #635, opened 2026-02-18
Allow user to discard all remaining optional effects

hooperk:
Provides the option to choose not to activate all remaining optional effects when only optional effects remain to be chosen from.

Useful for certain decks that have a lot of identical optional effects trigger at once but only need to activate 1, like Xros Heart or Eaters.

https://discord.com/channels/839041812932329512/1454784427019210887/1454784427019210887

I'll be honest, I'm not too pushed on this but it was a small enough change to rescue from the old repo, and it would still be nice to have.


x89rt2:
I did a basic check on this and it seemed to work fine. It will mean that we will have to stop making the optional effects false to reduce clicks but maybe we can be smart about which we choose to be false or true.

x89rt2 (approved):

PR #636, opened 2026-02-18
ST21-09 Unregister OPT use if no processing occured

hooperk:
ST21-09 as the proof so we can test it then this can be applied to many more cards that need the option to reset OPT if no processing occurred


x89rt2:
I don't remember if I asked last time, but in the off chance that we get twice per turns working again, and 1 is used up, and a 2nd one is used but then canceled, would this code support still having 1 opt used up?

hooperk:
> I don't remember if I asked last time, but in the off chance that we get twice per turns working again, and 1 is used up, and a 2nd one is used but then canceled, would this code support still having 1 opt used up?

It would. This removes only the last one use of the method, as each usage is unique. This is actually why we need to use hashStrings to compare usually, since the actual objects of each activation are considered different

hooperk:
Method does not work, looking instead at determining IsOptional with a function

PR #685, opened 2026-02-20
Implement AD1-024 Imperialdramon: FM

### LINKED ISSUE #665: AD1-024 Imperialdramon: Fighter Mode
Image

hooperk:


Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
```diff
@@ -0,0 +1,229 @@
...
+            string SharedEffectName = "Bottom deck 1 Opponent's lowest DP Digimon";
+
+            string SharedEffectDescription(string tag) => $"[{tag}] [Once Per Turn] Return 1 of your opponent's lowest DP Digimon to the bottom of the deck.";
+
+            bool IsLowestDPEnemy(Permanent permanent) => CardEffectCommons.IsMinDP(permanent, card.Owner.Enemy);
+
+            bool SharedCanActivateCondition(Hashtable hashtable)
+            {
+                return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
+                    && CardEffectCommons.HasMatchConditionOpponentsPermanent(card, IsLowestDPEnemy);
```
x89rt2: needs to always activate

Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
```diff
@@ -0,0 +1,229 @@
...
+                        && targetPermanent.TopCard.EqualsTraits("Hero");
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 5, ignoreDigivolutionRequirement: false, card: card, condition: null));
+            }
+            if (timing == EffectTiming.None)
+            {
+                static bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.EqualsCardName("Imperialdramon: Dragon Mode") || targetPermanent.TopCard.EqualsCardName("Imperialdramon Dragon Mode");
```
x89rt2, line 30: I'm 99% sure we fixed all the ones without the colon but we can leave if you want

Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
```diff
@@ -0,0 +1,229 @@
...
+
+                        bool unsuspend = GManager.instance.userSelectionManager.SelectedBoolValue;
+
+                        if (unsuspend)
+                        {
+                            yield return ContinuousController.instance.StartCoroutine(new IUnsuspendPermanents(
+                                new List<Permanent>() { card.PermanentOfThisCard() },
+                                activateClass).Unsuspend());
+                        }
+                    }
```
x89rt2, line 202: This seems like a convoluted way to do this, Just have an if opponent has digimon have the suspend their selector do it's thing, then outside of that if just select if unsuspend self

x89rt2 (changes requested):

Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
hooperk: The reason for this is because the "may" is attached to both, so if they can suspend an opponent and don't, they aren't allowed to unsuspend

x89rt2: Ah, you are right.

Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
hooperk: I copied the older cards but kinda better safe than sorry

x89rt2 (approved):

PR #690, opened 2026-02-20
Update EX11_066.cs

x89rt2:
Next attempt at fixing softlock that happens when trigger digimon leaves battle area before Xeno AT activation..

### DIFF
--- a/Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
+++ b/Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
@@ -198,7 +198,7 @@ namespace DCGO.CardEffects.EX11
                             yield return null;
                         }
 
-                        if (selectedCards.Count > 0)
+                        if (selectedCards.Count > 0 && CardEffectCommons.HasMatchConditionOwnersPermanent(card, targetPermanents.Contains))
                         {
                             while (selectedCards.Any())
                             {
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
```diff
@@ -232,59 +232,57 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
                                 }
                                 else selectedPermament = targetPermanents[0];
 
-                                if (selectedPermament != null)
+                                List<CardSource> digivolutionCards_fixed = new List<CardSource>();
+                                digivolutionCards_fixed.Clear(); // Clear here in cases of multiple loops
```
hooperk, line 236: Clear is unneccessary here because you just declared it as a brand new list.

x89rt2: That was put there by Lewis himself when he did the loop aspect of this card. Does it hurt anything leaving it in?

hooperk (approved):

Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
hooperk: No, and I have approved since it overall looks fine

PR #691, opened 2026-02-20
Implement different method for self deleting at turn end

hooperk:
Attempt to put the effect to delete on the permanent itself but not on any specific card in the stack.
If this works, the approach could be useful for many other effects that should effect the stack, such as immunity to digimon effect which currently is lost if de-digivolved.

This effect should now attempt to delete the Digimon every time the valid timing occurs, as the rulings state they should, without slowing the game down with player effects that never get cleared.

This can be tested by digivolving the played card into something with Armor Purge.

### DIFF
--- a/Assets/Scripts/CardEffect/BT23/Blue/BT23_017.cs
+++ b/Assets/Scripts/CardEffect/BT23/Blue/BT23_017.cs
@@ -235,67 +235,8 @@ namespace DCGO.CardEffects.BT23
 
                                 #region Delete Digimon Played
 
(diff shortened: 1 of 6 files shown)


x89rt2:
Best way to test this is with one of the CS rookies. We can use Eater Bit to protect it for various turns.

hooperk:
I'll implement this onto the Cs rookies in a bit

hooperk:
Not working for CS rookies locally, continuing to look into it

x89rt2 (approved):

PR #695, opened 2026-02-22
Fix turn passing during attack and execute deleting at wrong step

hooperk:
Tested by attacking with execute, delete was at correct time, and the turn did not pass so my attacks correctly checked opponent's security and not my own.

### DIFF
--- a/Assets/Scripts/Script/CardEffectFactory.cs
+++ b/Assets/Scripts/Script/CardEffectFactory.cs
@@ -715,4 +715,44 @@ public partial class CardEffectFactory
     }
 
     #endregion
+
+    #region Effect of a Permanent to Delete Itself
+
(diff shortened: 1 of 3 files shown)


Assets/Scripts/Script/CardEffectFactory.cs
```diff
@@ -715,4 +715,44 @@ public static IEnumerator ReplaceTopSecurityWithFaceUpOptionEffect(CardSource ca
     }
 
     #endregion
+
+    #region Effect of a Permanent to Delete Itself
+
+    public static ActivateClass DeleteSelfEffect(Permanent permanent, bool deleteOnOwnturn = true, bool deleteOnOpponentsTurn = true)
```
hooperk, line 721: FYI: Brought this forward from the rework to "delete at turn end" I am intending, uses the new pattern of being an effect of the permanent itself so that if it digivolves or de-digivolves it will still have the effect on itself

x89rt2 (approved):

PR #700, opened 2026-02-22
Update EX11_066.cs

x89rt2:
Maybe something like this?

### DIFF
--- a/Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
+++ b/Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
@@ -176,115 +176,133 @@ namespace DCGO.CardEffects.EX11
 
                         List<CardSource> selectedCards = new List<CardSource>();
 
-                        yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.SimplifiedRevealDeckTopCardsAndSelect(
-                            revealCount: 2,
-                            simplifiedSelectCardConditions:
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
```diff
@@ -176,116 +176,134 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
 
                         List<CardSource> selectedCards = new List<CardSource>();
 
-                        yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.SimplifiedRevealDeckTopCardsAndSelect(
-                            revealCount: 2,
-                            simplifiedSelectCardConditions:
-                            new SimplifiedSelectCardConditionClass[]
+                        if (CardEffectCommons.HasMatchConditionOwnersPermanent(card, targetPermanents.Contains))
```
hooperk, line 179: This will mean the cards are not revealed and still at top of deck if no valid targets, that isn't correct

hooperk (changes requested):

Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
hooperk: NVM, missed you had the else at the bottom, which I was about to suggest as an option. Yeah this'll work

hooperk (approved):

PR #705, opened 2026-02-23
Fix bug where multiple deletions cause this to miss the timing

screx:
## Summary
- Fix bug where only one [On Deletion] effect triggers when multiple Petrification Tokens are deleted simultaneously

## Problem
When multiple Petrification Tokens were deleted at the same time (e.g., from a board wipe effect), only one "Trash your top security card" effect would trigger instead of one per token.

## Root Cause
The `CanUseCondition` for the Petrification Token's [On Deletion] effect contained an unnecessary `IsExistOnBattleAreaDigimon(card)` check that isn't used by other [On Deletion] effects. This check could fail for subsequent tokens during batch deletion processing.

## Solution
Removed the redundant `IsExistOnBattleAreaDigimon(card)` check from the trigger condition, aligning with the standard pattern used by other [On Deletion] effects (e.g., Familiar Token, ST6_01).

**Before:**
```csharp
{
    if (CardEffectCommons.IsExistOnBattleAreaDigimon(card))
    {
        if (CardEffectCommons.CanTriggerOnDeletion(hashtable, card))
        {
            return true;
        }
    }
    return false;
}
```

**After:**
```csharp
{
    return CardEffectCommons.CanTriggerOnDeletion(hashtable, card);
}
```


x89rt2:
https://discord.com/channels/839041812932329512/1475565513223704699

@hooperk

screx:
still trying to recreate this but its only happened against my friend so far RIP

hooperk (approved):
While I am not 100% sure on the scenario itself, this change seems a correct one to make to clean up the code

PR #706, opened 2026-02-25
Update EX11_066.cs

x89rt2:
Ran into one of the Vemmon not being selectable issue once in version 1.14.3. It's damn near impossible to consistently recreate though. Since doing the normal method is bringing up additional dialogue boxes, I'm using the fancy version again. I did narrow down the problem to the selection that starts on line 240 or so of this code. As the next test for that, I'll be removing the clear code as will be shown in the next commit.

### DIFF
--- a/Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
+++ b/Assets/Scripts/CardEffect/EX11/Black/EX11_066.cs
@@ -293,7 +293,7 @@ namespace DCGO.CardEffects.EX11
                             revealCount: 2,
                             simplifiedSelectCardCondition:
                             new SimplifiedSelectCardConditionClass(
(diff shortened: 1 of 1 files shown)


x89rt2:
shit, it's softlocking on played against ruin mode again

x89rt2:
issue starts on line 292 somewhere in that else

x89rt2:
A little ugly but primary testing in open stopped having a soft lock. Let me know if there's a better method cause null on cantargetcondition was the cause of the softlock.

x89rt2:
I wonder if there's a problem with the cantargetcondition that causes it to sometimes fail leading to the other card bug.

hooperk (approved):

PR #719, opened 2026-02-27
Fix gaining immunity should prevent deletion

hooperk:
If after immediate effects the Digimon now has immunity to deletion, it should not be deleted, but CanBeDestroyedBySkill was not being checked again.

Example scenario: Lucemon: Satan Mode with Lucemon: Larva mode in breeding. Satan mode is reduced to 0 DP and would be deleted. as an immediate effect, Larva Mode is raised and it's All turns starts applying: 0 DP Lucemon Digimon may not be deleted.

In progress DestroyPermanentsClass.Destroy() was proceeding to the deletion as Larva mode had not explicitly removed Satan Mode from deletion

### DIFF
--- a/Assets/Scripts/Script/CardController.cs
+++ b/Assets/Scripts/Script/CardController.cs
@@ -3472,7 +3472,8 @@ public class DestroyPermanentsClass
         List<Permanent> destroyTargetPermanents_Fixed = _destroytargetPermanents.Filter(permanent =>
             permanent != null
             && permanent.TopCard != null
-            && permanent.willBeRemoveField);
+            && permanent.willBeRemoveField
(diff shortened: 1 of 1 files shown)


x89rt2:
This doesn't seem correct. The reason in that situation why the deletion is prevented is because larva's effect specifically says "they don't leave". This code is giving me gallant with wargrowl ace and protoform vibes which we know does not work like that,

hooperk:
We believe https://github.com/DCGO2/DCGO/pull/720 should actually address the root cause of what is happening here. Will test both scenarios with that one PR first before trying this change

x89rt2:
Larva seems good according to beta so I'm closing this.

PR #720, opened 2026-02-27
Fix 0 DP Digimon being deleted in breeding

### LINKED ISSUE #522: BT18-086 Lucemon: Larva
Fails to protect Satan from DP deletion when is moved out of breeding area by it's own effect to protect- same as middle bug
https://discord.com/channels/839041812932329512/1189361807727542393/1309895066831032463
https://discord.com/channels/839041812932329512/1310293820583182347
Deleted in raising when something dies to dp deletion
https://discord.com/channels/839041812932329512/1309916056856301629
Protecting opponent's 0 dp digimon
https://discord.com/channels/839041812932329512/1310313387250548796

hooperk:
The check for if 0 DP Digimon should be deleted correctly checks only battle area digimon, but if such a deletion occurs the actual process deletes 0 DP Digimon in the breeding area as well.

Example: Lucemon: Larva Mode and any Digimon in the battle area being deleted for having 0 DP.


x89rt2:
approved but I think I want to get current build into public before merging

x89rt2 (approved):

PR #729, opened 2026-02-28
Implement P-237 and Use Requirements

### LINKED ISSUE #727: P-237 Unique Emblem: Machina's Ascension
Image

hooperk:
Made Use Requirements a CEF for 2 reasons: 1, it's going to be super common keyword and 2: We are still waiting on confirmation from Bandai on rulings for it (like if it counts breeding) and it'll be easier to adjust this to those rulings if it is in a central location

### DIFF
--- a/Assets/Scripts/CardEffect/P/Green/P_237.cs
+++ b/Assets/Scripts/CardEffect/P/Green/P_237.cs
@@ -1,0 +1,283 @@
+using System.Collections;
+using System.Collections.Generic;
+using System;
+
+//Unique Emblem: Machina's Ascension
+namespace DCGO.CardEffects.P
+{
+    public class P_237 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Ignore Color Requirment
+            if (timing == EffectTiming.None)
+            {
+                cardEffects.Add(CardEffectFactory.UseRequirements(card, HasMaquinamonInText));
+
+                bool HasMaquinamonInText(Permanent permanent)
+                {
+                    return permanent.TopCard.HasText("Maquinamon");
+                }
+            }
+            #endregion
+
+            #region Main
+            if (timing == EffectTiming.OptionSkill)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Play 1 [Maquinamon]/[Unchained] from hand or trash, then place in battle area", CanUseCondition, card);
+                activateClass.SetUpActivateClass(null, ActivateCoroutine, -1, false, EffectDiscription());
+                cardEffects.Add(activateClass);
+
+                string EffectDiscription()
+                {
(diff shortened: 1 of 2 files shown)


Assets/Scripts/CardEffect/P/Green/P_237.cs
```diff
@@ -0,0 +1,283 @@
...
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerOptionMainEffect(hashtable, card);
+                }
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                {
+                    return cardSource.EqualsCardName("Maquinamon") || cardSource.EqualsCardName("Unchained")
```
x89rt2:
```suggestion
                    return (cardSource.EqualsCardName("Maquinamon") || cardSource.EqualsCardName("Unchained"))
```

Assets/Scripts/CardEffect/P/Green/P_237.cs
```diff
@@ -0,0 +1,283 @@
...
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card);
+                }
+
+                bool PlayedPermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card)
+                        && permanent.TopCard.EqualsCardName("Unchained");
+                }
```
x89rt2, line 187: Correct me if I'm wrong, but this only checks that it's in battle area in trigger timing, correct? If between Unchained being played and this Delay getting activated, Unchained gets deleted, the Delay can still activate with this code, no?

x89rt2 (changes requested):

Assets/Scripts/CardEffect/P/Green/P_237.cs
hooperk: Unnecessary, there is no mixture of || or && here. Visual studio even warns it is unnecessary by greying them out

Assets/Scripts/CardEffect/P/Green/P_237.cs
hooperk: Correct. Unchained being played is the trigger condition but is not a processing condition

Assets/Scripts/CardEffect/P/Green/P_237.cs
x89rt2: There is a 
                        && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
afterwards that is only checking if the unchained can be played and not the maquina. I probably should have grabbed it in the first review, but eh.

x89rt2:
Github is saying there are conflicts with the cardeffectfactory. I'm bad at fixing those so that will have to be you.

Assets/Scripts/CardEffect/P/Green/P_237.cs
hooperk: Woops, couldn't see that when I double checked but you're right

hooperk:
Going to mark this one as ready to review, it technically should be the first card with Use Reqs so would be good to have that in testing

x89rt2 (approved):

PR #730, opened 2026-03-01
Vortexdramon with Digimon Immunity as a Permanent Effect, instead of card effect

hooperk:
PermanentEffectFactory for given effects on Permanents, which then should continue to work correctly through de-digivolution as long as the permanent remains for the effect given duration.

Having tested the pattern works on Execute granting a given self delete effect, next test will be on Vortexdramon's digimon immunity. Once proofed there we can copy the pattern and apply to other effects being granted to the entire stack.

### DIFF
--- a/Assets/Scripts/CardEffect/EX11/Green/EX11_074.cs
+++ b/Assets/Scripts/CardEffect/EX11/Green/EX11_074.cs
@@ -124,31 +124,13 @@ namespace DCGO.CardEffects.EX11
 
                 if (ownDigimon)
                 {
+                    Permanent thisPermanent = card.PermanentOfThisCard();
+
                     #region Give Digimon Effect Immunity
 
-                    bool CanUseCondition1(Hashtable hashtable)
-                    {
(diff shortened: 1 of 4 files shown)


x89rt2:
This code reminds me that we have an issue with ex10 apoca. The effect to play out 3 guys and deletes them at end of turn is supposed to happen all at same time and the source should be from apoca (regardless of if he's still in battle area or not), but it's currently behaving as each of the digimon's end of turn delete themselves letting them go 1 by 1 and triggering their on deletions before the other ones are deleted.

hooperk:
> This code reminds me that we have an issue with ex10 apoca. The effect to play out 3 guys and deletes them at end of turn is supposed to happen all at same time and the source should be from apoca (regardless of if he's still in battle area or not), but it's currently behaving as each of the digimon's end of turn delete themselves letting them go 1 by 1 and triggering their on deletions before the other ones are deleted.

That would be a fairly unique effect... it could be made though. Would need to be an effect on the player. One question then: should it attempt to delete every end of turn forever or only the first? I assume every

x89rt2 (approved):

PR #744, opened 2026-03-05
Simplify CanActivateCondition logic

x89rt2:
Fix ESS not trigger if oppo had 0 cards in sec.

### DIFF
--- a/Assets/Scripts/CardEffect/EX8/Black/EX8_052.cs
+++ b/Assets/Scripts/CardEffect/EX8/Black/EX8_052.cs
@@ -294,8 +294,7 @@ namespace DCGO.CardEffects.EX8
 
                 bool CanActivateCondition(Hashtable hashtable)
                 {
-                    return card.Owner.Enemy.SecurityCards.Count > 0 &&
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/EX8/Black/EX8_052.cs
```diff
@@ -294,8 +294,7 @@ bool CanUseCondition(Hashtable hashtable)
 
                 bool CanActivateCondition(Hashtable hashtable)
                 {
-                    return card.Owner.Enemy.SecurityCards.Count > 0 &&
```
hooperk, line 297: Can't check myself right now, do we need to add this check inside the coroutine anywhere?

x89rt2: Unless

yield return ContinuousController.instance.StartCoroutine(new IDestroySecurity

needs it, then no.

hooperk: It checks itself internally so all good

hooperk (approved):

PR #749, opened 2026-03-06
Fix EX10-061 played permanents being deleted individually

hooperk:
Effect should now:
- Delete the played Digimon simultaneously.
- Continue to do so every turn end until all permanents played by the effect are deleted. It will then stop triggering every end step.
- Appear as an effect of the Apocalymon card.

### DIFF
--- a/Assets/Scripts/CardEffect/EX10/White/EX10_061.cs
+++ b/Assets/Scripts/CardEffect/EX10/White/EX10_061.cs
@@ -395,201 +395,194 @@ namespace DCGO.CardEffects.EX10
 
             #endregion
 
+            #region Shared OP / WD
+
+            string SharedEffectName = "Play 1 [Dark Masters] from sources, give <Rush>, Delete them at end of turn";
+
+            string SharedEffectDescription(string tag)
+                =>  $"[{tag}] You may play 1 of each [Dark Masters] trait card with different names from this Digimon's digivolution cards without paying the costs. Then, all of your [Dark Masters] trait Digimon gain <Rush> for the turn. At turn end, delete the Digimon this effect played.";
+
+            bool SharedCanActivateCondition(Hashtable hashtable) => CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/EX10/White/EX10_061.cs
```diff
@@ -395,401 +395,211 @@ bool isUpDown()
...
-                                    if (!playedPermanent.TopCard.CanNotBeAffected(activateClass))
-                                    {
-                                        return true;
-                                    }
-                                }
+                    ActivateClass activateClass1 = new ActivateClass();
 
-                                return false;
-                            }
+                    Func<EffectTiming, ICardEffect> cardEffectFunction = GetCardEffect;
```
hooperk, line 512: The unusual ordering of these lines is due to dependencies. ActivateCoroutine needs the function to be assigned which needs the activate class to have been created.

Assets/Scripts/CardEffect/EX10/White/EX10_061.cs
```diff
@@ -395,401 +395,212 @@ bool isUpDown()
...
-                                    selectPlayer: card.Owner,
-                                    cardEffect: activateClass);
-
-                        selectCardEffect.SetUpCustomMessage("Select 1 digivolution card to play.", "The opponent is selecting 1 digivolution card to play.");
-                        selectCardEffect.SetUpCustomMessage_ShowCard("Played Card");
-
-                        yield return StartCoroutine(selectCardEffect.Activate());
-
-                        IEnumerator SelectCardCoroutine(CardSource cardSource)
+                        if (playedPermanents.Count() <= 0)//Remove effect from activating every turn once all permanents deleted
```
x89rt2, line 551: Is this checking battle area?

hooperk: Yes. I don't believe there's any effects to move a Digimon that is already in play to breeding so it should be ok we aren't checking breeding area

x89rt2 (approved):

PR #760, opened 2026-03-09
Bulk: Show pendign effects when looking at digivolution cards

hooperk:
Once we're satisfied the general fix works as expected, here's several more candidates to ensure they display correctly


Assets/Scripts/CardEffect/BT23/Yellow/BT23_033.cs
```diff
@@ -138,8 +138,8 @@ IEnumerator SelectCardCoroutine(CardSource cardSource)
                             canEndNotMax: false,
                             isShowOpponent: true,
                             mode: SelectCardEffect.Mode.Custom,
-                            root: SelectCardEffect.Root.Custom,
-                            customRootCardList: thisPermament.StackCards,
```
x89rt2, line 142: this stackcards is interesting. anything noteworthy on it?

Assets/Scripts/CardEffect/BT21/Red/BT21_023.cs
```diff
@@ -327,7 +327,7 @@ IEnumerator SharedLinkedActivateCoroutine(Hashtable hashtable, ActivateClass act
 
                 string EffectDiscription()
                 {
-                    return "[Your Turn] [Once Per Turn] When this Digimon gets linked, delete 1 of your opponent�s Digimon with as much or less DP as this Digimon";
+                    return "[Your Turn] [Once Per Turn] When this Digimon gets linked, delete 1 of your opponent�s Digimon with as much or less DP as this Digimon";
                 }
```
x89rt2, line 331: let's ignore the misspelled discription but let's at least fix that weird symbol.

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT23/Yellow/BT23_033.cs
hooperk: StackCards includes the TopCard itself as well as digivolution cards, while excluding link cards. This isn't actually a problem since the top cards involved don't have a link condition, but I'd like to make the code that little bit more correct while I see it

x89rt2 (approved):

PR #764, opened 2026-03-10
Update BT15_084.cs

x89rt2:
We want to give this a shot?

### DIFF
--- a/Assets/Scripts/CardEffect/BT15/Yellow/BT15_084.cs
+++ b/Assets/Scripts/CardEffect/BT15/Yellow/BT15_084.cs
@@ -123,12 +123,12 @@ namespace DCGO.CardEffects.BT15
                     {
                         if (CardEffectCommons.CanTriggerWhenLoseSecurity(hashtable, player => player == card.Owner))
                         {
-                            ICardEffect effect = CardEffectCommons.GetCardEffectFromHashtable(hashtable);
-                            if (effect != null)
-                            {
-                                return true;
-                            }
-                            //if (!GManager.instance.attackProcess.DoSecurityCheck)
+                            //ICardEffect effect = CardEffectCommons.GetCardEffectFromHashtable(hashtable);
+                            //if (effect != null)
+                            //{
+                            //    return true;
+                            //}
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT15/Yellow/BT15_084.cs
```diff
@@ -123,12 +123,12 @@ bool CanUseCondition(Hashtable hashtable)
...
-                            {
-                                return true;
-                            }
-                            //if (!GManager.instance.attackProcess.DoSecurityCheck)
+                            //ICardEffect effect = CardEffectCommons.GetCardEffectFromHashtable(hashtable);
+                            //if (effect != null)
+                            //{
+                            //    return true;
+                            //}
+                            return !GManager.instance.attackProcess.DoSecurityCheck;
```
hooperk, line 131: Hmm... yeah that looks like it would effectively cover all current non-effect reasons for security to be trashed, which is currently just battle. Will cause an issue if there's ever Overflow<trash 1 sec>...

Was checking the effect was null not working somehow? Were there effects not correctly including the card effect that caused the trash?

x89rt2: there's a long list in the internal doc, although I don't know how recent those are

x89rt2: at least 2 were just confirmed so I think it's worth a shot maybe.

hooperk: The thing is there is a good chance the problem is actually with the cards that don't trigger it. This is written correctly and so is going to match a lot of other effects. In fact I can see a bug in some shared code, lemme put that up and confirm how many cars it should fix

hooperk: I have an idea how to track down everywhere that is not correctly triggering this effect, and even make sure we don't make the mistake again in the future. Will apply it after work today

x89rt2:
Unnecessary

PR #767, opened 2026-03-11
Fix remove default parameter to ensure it is always correctly set

hooperk:
Initially removed the default as a way to track down all files with this bug, but given there is exactly 1 place where a null is correct, it is better to just always expect the parameter and pass a null in that one case. As such, I have removed the default for the cardEffect parameter. This prevents us from recreating the bug in the future.

### DIFF
--- a/Assets/Scripts/Script/CardController.cs
+++ b/Assets/Scripts/Script/CardController.cs
@@ -3710,7 +3710,8 @@ public class ISecurityCheck
 
                         yield return ContinuousController.instance.StartCoroutine(new IReduceSecurity(
                             player: brokenSecurityCard.Owner,
-                            refSkillInfos: ref triggeredSkillInfos).ReduceSecurity());
+                            refSkillInfos: ref triggeredSkillInfos,
+                            null).ReduceSecurity());
 
                         #region security effect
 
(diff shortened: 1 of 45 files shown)


x89rt2:
We can try that if you want, but I'm told a promo patamon did not trigger it and I'm not seeing that in the list.

https://discord.com/channels/1095717982522585098/1392137159275708456/1481075531087613994

hooperk:
@x89rt2 you're right, I started on something to get every other card that is a problem and then had to head to work. Hoping to finish tonight. Will mark as draft for now

Assets/Scripts/Script/CardController.cs
```diff
@@ -3794,7 +3710,8 @@ bool StopSecurityCheck()
 
                         yield return ContinuousController.instance.StartCoroutine(new IReduceSecurity(
                             player: brokenSecurityCard.Owner,
-                            refSkillInfos: ref triggeredSkillInfos).ReduceSecurity());
+                            refSkillInfos: ref triggeredSkillInfos,
+                            null).ReduceSecurity());
```
hooperk, line 3714: FYI: This is part of the game's process for performing a security check per battle and is the one place that does not have a card effect, so null is correct

x89rt2 (approved):

PR #770, opened 2026-03-11
Fix Ending attack with effect puts End Of Attack out of order

hooperk:
When an effect ends the attack, this should immediately trigger End of Attack timing, simultaneous to anything caused by the same effect.
Example:
EX9-024 Hanumon in Digivolution cards, deleting a Puppet token to end the attack.
Expected Result:
End of Attack effect of the attacking digimon has turn player priority and goes before the -3k from the puppet
Actual Result:
End of attack effects are waiting until all current processing is completed to even trigger. Thus the -3K gets to activate first.

Similar to AttackProcess.Attack() being used to start an attack, AttackProcess.EndAttack() Ends the attack, triggering EndOfAttack and setting up that the attack will end after all pending effects are resolved, at which point cleanup will run to actually end the attack, thus ensuring that any End Of attack effects still process with an ongoing attack.

### DIFF
--- a/Assets/Scripts/CardEffect/BT13/Purple/BT13_088.cs
+++ b/Assets/Scripts/CardEffect/BT13/Purple/BT13_088.cs
@@ -428,9 +428,9 @@ namespace DCGO.CardEffects.BT13
                             mode: SelectHandEffect.Mode.Discard,
                             cardEffect: activateClass);
 
-                        yield return StartCoroutine(selectHandEffect.Activate());
+                        yield return ContinuousController.instance.StartCoroutine(selectHandEffect.Activate());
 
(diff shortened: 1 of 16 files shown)


x89rt2:
Was this an issue? I don't remember receiving any complaints about this.

hooperk:
> Was this an issue? I don't remember receiving any complaints about this.

@x89rt2 It's something I saw come up in the rulings discord and immediately knew that it was not programmed to handle correctly. Has a pretty straightforward testing scenario though, and I was able to confirm attacks seemed to still work properly without being stopped, although I cannot test the specific end attack scenarios on my local

x89rt2 (approved):

PR #771, opened 2026-03-11
Implement AD1-005 Gaiamon ACE

### LINKED ISSUE #645: AD1-005 Gaiamon
Image

hooperk:


Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            #region Alternative Digivolution Condition - Ult.
+            if (timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.EqualsTraits("Ult.");
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 5, ignoreDigivolutionRequirement: false, card: card, condition: null));
```
x89rt2:
```suggestion
                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 4, ignoreDigivolutionRequirement: false, card: card, condition: null));
```

Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
```diff
@@ -0,0 +1,274 @@
...
+                    {
+                        selectionElements.Add(new (message: $"Link from Hand", value : 1, spriteIndex: 0));
+                    }
+                    if (validDigivolutionCardCount > 0)
+                    {
+                        selectionElements.Add(new (message: $"Link from Digivolution Cards", value : 2, spriteIndex: 0));
+                    }
+                    selectionElements.Add( new (message: (selectedCards.Count == 0 ? $"Don't link" : $"Finish Linking"), value: 3, spriteIndex: 1));
+
+                    string selectPlayerMessage = "From which area will you link a Maquinamon?";
```
x89rt2: This seems interesting, talk to me about it:

(selectedCards.Count == 0 ? $"Don't link" : $"Finish Linking")

No maquinas are linked with the making of this effect.

Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
```diff
@@ -0,0 +1,274 @@
...
+                                        selectCardCoroutine: SelectCardCoroutine,
+                                        afterSelectCardCoroutine: null,
+                                        message: "Select cards to link.",
+                                        maxCount: maxCount,
+                                        canEndNotMax: true,
+                                        isShowOpponent: true,
+                                        mode: SelectCardEffect.Mode.Custom,
+                                        root: SelectCardEffect.Root.DigivolutionCards,
+                                        customRootCardList: thisPermanent.DigivolutionCards,
+                                        canLookReverseCard: true,
```
x89rt2: remind me, with canlookreverse, we can look but no touch, correct?

Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
```diff
@@ -0,0 +1,274 @@
...
+                    List<SelectionElement<int>> selectionElements = new List<SelectionElement<int>>();
+                    if (validHandCardCount > 0)
+                    {
+                        selectionElements.Add(new (message: $"Link from Hand", value : 1, spriteIndex: 0));
+                    }
+                    if (validDigivolutionCardCount > 0)
+                    {
+                        selectionElements.Add(new (message: $"Link from Digivolution Cards", value : 2, spriteIndex: 0));
+                    }
+                    selectionElements.Add( new (message: (selectedCards.Count == 0 ? $"Don't link" : $"Finish Linking"), value: 3, spriteIndex: 1));
```
x89rt2, line 117: If we want to only link 1 but have 2 or more available, does selecting don't link go to end loop?

x89rt2: Is that what the default on line 182 is?

x89rt2 (changes requested):

Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
hooperk: Oh woops, didn't change after copy, will update. The first part determines the text to display for the stop button. If they've not selected any cards yet, "don't link" but if they have chosen 1 card "Finish Linking".

Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
hooperk: It allows both look and select, and you need to filter out in the can select card condition when it should not be able to select but can see

Assets/Scripts/CardEffect/AD1/Red/AD1_005.cs
hooperk: You'll need to select 1, end not max and then select "Finish Linking" This is because we also need to allow a way for you to select 1 of 2 valid cards from hand and then 1 of 2 valid cars from digivolution cards or vice versa

hooperk: The default means if you pick anything other than 1 or 2 (so 3, which is don't link / finish linking), exit the loop and stop looking to add cards

x89rt2 (approved):

PR #773, opened 2026-03-12
Fix piercing checks not done if defender is removed after activation

hooperk:
Step was ending early if defender was removed, however if piercing had already activated the correct behaviour is that the security check should still be performed after the lack of battle


x89rt2:
I'm seeing a lot of red but little green. Is the lower deleted code just old remnant code that all your new changes made superfluous?

hooperk:
> I'm seeing a lot of red but little green. Is the lower deleted code just old remnant code that all your new changes made superfluous?

I deleted a block of commented out very old code. I think early on the devs had been looking at how to support a battle by effect, or may have meant the full attack by effect. either way, both are sorted out now.

x89rt2 (approved):

PR #782, opened 2026-03-14
Implement Mode.Degenerate with AD1-018 to test

hooperk:
Saw an opportunity for another Mode to reduce the code we have to write which we can test on one of the new cards.

Called "Degenerate" to match to existing method, and as that method is not just for De-digivolve but also other effects that remove the top digivolution card of a stack.


x89rt2:
> Saw an opportunity for another Mode to reduce the code we have to write which we can test on one of the new cards.
> 
> Called "Degenerate" to match to existing method, and as that method is not just for De-digivolve but also other effects that remove the top digivolution card of a stack.

Keep in mind some effects prevent de-digi but do not prevent the shoutmon top card strip.

x89rt2 (approved):

PR #790, opened 2026-03-15
Update AD1_024.cs

x89rt2:
Triggers from egg hatching.

### DIFF
--- a/Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
+++ b/Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
@@ -140,7 +140,11 @@ namespace DCGO.CardEffects.AD1
                             || CardEffectCommons.CanTriggerWhenPermanentDigivolving(hashtable, IsDigimonCondition));
                 }
 
-                bool IsDigimonCondition(Permanent permanent) => permanent.IsDigimon;
+                bool IsDigimonCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && permanent.IsDigimon;
+                }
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
```diff
@@ -140,7 +140,11 @@ bool CanUseCondition(Hashtable hashtable)
                             || CardEffectCommons.CanTriggerWhenPermanentDigivolving(hashtable, IsDigimonCondition));
                 }
 
-                bool IsDigimonCondition(Permanent permanent) => permanent.IsDigimon;
+                bool IsDigimonCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && permanent.IsDigimon;
+                }
```
hooperk:
```suggestion
                bool IsDigimonCondition(Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card);
```

hooperk (changes requested):

Assets/Scripts/CardEffect/AD1/Blue/AD1_024.cs
x89rt2: no cause it also triggers from opponent digimon played

hooperk: Oh, then IsPermanentExistsOnOpponentBattleAreaDigimon.

The important thing is that `CardEffectCommons.IsExistOnBattleArea(card)` only ever tells us the state of the card itself, The Imperialdramon. In a permanent condition you need to use one of these methods that checks the status of the permanent in question

hooperk (approved):

PR #814, opened 2026-03-16
Create AD1_023.cs

### LINKED ISSUE #664: AD1-023 J.P. Koji, Koichi
Image

x89rt2:


Assets/Scripts/CardEffect/AD1/Black/AD1_023.cs
```diff
@@ -0,0 +1,287 @@
...
+
+                            yield return ContinuousController.instance.StartCoroutine(selectCardEffect.Activate());
+                            break;
+                        }
+                        default:
+                            goto END_LOOP;
+                    }
+
+                    if (selectedCards.Count == prevCount)
+                        break;
```
hooperk: Remove this, user should be allowed decide not to place from trash and still place from hand

hooperk (changes requested):

Assets/Scripts/CardEffect/AD1/Black/AD1_023.cs
x89rt2: exact same code from the other tamer so it will be an issue there as well

hooperk (approved):

PR #815, opened 2026-03-17
Fix AD1-021 not resetting OPT use

hooperk:


x89rt2:
@hooperk tested internally, did not work.

image

PR #830, opened 2026-03-17
Make effect optional only when can't do mandatory processing

hooperk:
Not the ideal solution, but is a way we can ensure correct behaviour for methods with mandatory and optional sections.

Did some testing. Was optional and not using OPT when not activated when no yellow agumon or greymon present, and was mandatory as soon as an agumon was present


x89rt2:
primary testing seemed good?

hooperk:
> primary testing seemed good?

Yup, expected behaviour with and without a yellow agumon present. Post AD1 we can look to apply this to Adventure cards

x89rt2 (approved):

PR #831, opened 2026-03-17
Fix Liberator 7s cannot warp in breeding

hooperk:


x89rt2:
I think if left like that then it will let you evo opponent's digimon.

hooperk:
> I think if left like that then it will let you evo opponent's digimon.

No alt condition ever checks who's battle area it is on and they work fine, such checks are handled elseswhere

x89rt2 (approved):

PR #838, opened 2026-03-17
Fix 3M trashed effects treated as Digimon Effects

hooperk:


Assets/Scripts/Script/ICardEffect.cs
```diff
@@ -35,22 +35,10 @@ public void SetUpICardEffect(string effectName, Func<Hashtable, bool> canUseCond
...
-            if (IsInheritedEffect || IsLinkedEffect)
-            {
-                SetIsDigimonEffect(true);
-            }
-            else if (card != null && CardEffectCommons.IsExistOnBattleArea(card))
-            {
-                SetIsDigimonEffect(card.PermanentOfThisCard().IsDigimon);
-                SetIsTamerEffect(card.PermanentOfThisCard().IsTamer);
-            }
-        }
```
hooperk, line 53: I am PRETTY sure this code is unnecessary with what is done elsewhere, and I think it will cause issues to leave this in. Removing it, but we will need a quick recheck of Koichi just to be sure I'm not breaking things.

x89rt2 (approved):

PR #839, opened 2026-03-17
Update BT22_013.cs

x89rt2:
Hand digivolution behaving as optional when mandatory.


Assets/Scripts/CardEffect/BT22/Red/BT22_013.cs
hooperk, line 119: also needs isOptional: false down here, and same on Melga

hooperk (changes requested):

hooperk (approved):

PR #845, opened 2026-03-18
Fix AD1-004 not getting DP increase without boost

hooperk:


Assets/Scripts/CardEffect/AD1/Red/AD1_004.cs
```diff
@@ -122,19 +122,8 @@ bool CanUseCondition(Hashtable hashtable)
 
             #region All Turns
 
-            if (timing != EffectTiming.None)
+            if (timing == EffectTiming.None)
```
hooperk, line 125: FYI: This was the issue. ChangeDP effects only count if they are in EffectTiming.None.
In comparison, the boost is adding or removing itself at every other timing in the game.

x89rt2 (approved):

PR #879, opened 2026-03-19
Update AD1_002.cs

x89rt2:
Fixed not called selectcardcoroutine


hooperk (approved):

hooperk (changes requested):
Bug listed on verification sheet does not make sense for this card, and listed changes do nothing

PR #885, opened 2026-03-19
Update BT5_109.cs

### LINKED ISSUE #1000: BT5-109 Mega Digimon Fusion!
Image

Permabanned card was never implemented. As we look into allowing banned list removal, we need to implement the card.

x89rt2:
Notes from previous reviews in old Repo:

hooperk reviewed on Jan 16
CardEffect/BT5/White/BT5_109.cs
                    {
                        if (targetPermanents != null)
                        {
                            if (targetPermanents.Count(PermanentCondition) >= 1)
Member
@hooperk
hooperk
on Jan 16
I think this should be ==1 in order to correctly be "when 1 of your level 6 digimon..."

Member
Author
@x89rt2
x89rt2
on Jan 16
Doesn't matter cause you can only do 1 digivolution at a time. Also, this "one" might be an "any".

Member
@hooperk
hooperk
on Jan 16
Jogress would count and is 2 permanents. Although currently there isn't a jogress with a non-0 cost for this to matter for, it does mean that jogressed digimon would be given the deck bounce effect despite not meeting this card's condition

Member
Author
@x89rt2
x89rt2
on Jan 16
• 
Am I still missing something or isn't that fine? since one means any? For the record, my brain is not really on right now. https://discord.com/channels/681578268729540663/749146995708395601/1461739750250647623

@x89rt2	Reply...
hooperk
hooperk reviewed on Jan 16
CardEffect/BT5/White/BT5_109.cs

                ICardEffect GetCardEffect(EffectTiming _timing)
                {
                    if (_timing == EffectTiming.BeforePayCost)
Member
@hooperk
hooperk
on Jan 16
If this effect goes off here, it will remove the cost reduction just before you would use int. Instead maybe this could go to when digivolving timing and remove the effect just after digivolution? EffectTiming.OnEnterFieldAnyone

@x89rt2	Reply...
hooperk
hooperk reviewed on Jan 16
CardEffect/BT5/White/BT5_109.cs

                IEnumerator ActivateCoroutine2(Hashtable _hashtable1)
                {
                    if (CardEffectCommons.CanTriggerWhenPermanentWouldDigivolve(
Member
@hooperk
hooperk
on Jan 16
If this does move to when digivolving timing, this will need to change with it

@x89rt2	Reply...
hooperk
hooperk reviewed on Jan 16
CardEffect/BT5/White/BT5_109.cs
                        cardCondition: null))
                    {
                        card.Owner.UntilEachTurnEndEffects.Remove(getCardEffect);
                        card.Owner.UntilEachTurnEndEffects.Remove(getCardEffect1);
Member
@hooperk
hooperk
on Jan 16
You should probably also add the deck bottom effect as part of this effect.

### DIFF
--- a/Assets/Scripts/CardEffect/BT5/White/BT5_109.cs
+++ b/Assets/Scripts/CardEffect/BT5/White/BT5_109.cs
@@ -30,11 +30,224 @@ public class BT5_109 : CEntity_Effect
 
             IEnumerator ActivateCoroutine(Hashtable _hashtable)
             {
-                Debug.Log("This card is banned.");
-                yield return null;
+                ContinuousController.instance.PlaySE(GManager.instance.GetComponent<Effects>().BuffSE);
+
+                yield return new WaitForSeconds(0.2f);
+
+                ActivateClass activateClass1 = new ActivateClass();
+                Func<EffectTiming, ICardEffect> getCardEffect = GetCardEffect;
+                activateClass1.SetUpICardEffect("Digivolution Cost -6", CanUseCondition1, card);
+                activateClass1.SetUpActivateClass(CanActivateCondition, ActivateCoroutine1, -1, true, EffectDiscription1());
+                CardEffectCommons.AddEffectToPlayer(effectDuration: EffectDuration.UntilEachTurnEnd, card: card, cardEffect: null, timing: EffectTiming.None, getCardEffect: getCardEffect);
+
+                ActivateClass activateClass2 = new ActivateClass();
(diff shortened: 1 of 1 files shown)


x89rt2:
Yea, this is beyond me. I'm dropping it. Anyone is free to pick it back up.

x89rt2:
Most current implementation has errors. I don't know of a way to make the digivolved digimon a permanent to stick the self bounce effect to it.

hooperk:
I think you had it most of the way there, I'll try finish it when I can make some time, but will leave low priority for now

hooperk (approved):

PR #887, opened 2026-03-19
Update AD1_023.cs

x89rt2:
fix color select issue

### DIFF
--- a/Assets/Scripts/CardEffect/AD1/Black/AD1_023.cs
+++ b/Assets/Scripts/CardEffect/AD1/Black/AD1_023.cs
@@ -1,4 +1,3 @@
-
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/AD1/Black/AD1_023.cs
```diff
@@ -68,11 +67,13 @@ IEnumerator SelectCardCoroutine(CardSource cardSource)
...
                 bool CanSelectCardCondition(CardSource cardSource)
                 {
+                    List<CardSource> concat = selectedCards.Concat(new List<CardSource>() { cardSource }).ToList();
+
                     if (!IsHybridCard(cardSource)) return false;
                     if (selectedCards.Contains(cardSource)) return false;
                     if (selectedCards.Count > 0)
                     {
-                        return cardSource.CardColors.Any(c => !selectedCards[0].CardColors.Contains(c));
+                        return Combinations.GetDifferenetColorCardCount(concat) >= concat.Count;
```
hooperk, line 76: This would prevent 2 different single colour cards. I have a fix for these

x89rt2: in that case, take a look at ex7 tlaclemon

PR #895, opened 2026-03-20
Fix Digixros not marked as Digixros

hooperk:


hooperk:
At some point I may need to put the japannese comments through google translate and replace with the english cause my machine keeps messing with it and I don't think any current dev can read it anyway

x89rt2 (approved):

PR #913, opened 2026-03-22
Fix Select X and Y effects taking 1 card as both

### LINKED ISSUE #480: Cards that search for "X and Y" are currently searching for "X, then Y"
[Bug report](https://discord.com/channels/1095717982522585098/1297669294208847882)

hooperk:
When an effect says to "select 1 X and 1 Y" from revealed cards, it is allowed to select a card that fulfills both as either condition when it would be the only card that could fulfill the other condition, thus taking only 1 digimon. For example on Davis Motomiya, if you have a green card, a blue green card and some other card that doesn't meet the conditions, it should be legal to select the blue green card "as the green card" and then not take another card. Previously you were forced to select the blue green card as your blue card, and then since you still have a valid green card must select that.
Now in this circumstance, the player is given the option to not select a card, thus taking the previously chosen card as their green card.

I have tested this with Davis specifically locally, all scenarios worked correctly.

Every (current) card which uses the relevant wording has been updated, although not all of them currently have a card that can fulfill both of their conditions.

### DIFF
--- a/Assets/Scripts/CardEffect/BT10/Green/BT10_046.cs
+++ b/Assets/Scripts/CardEffect/BT10/Green/BT10_046.cs
@@ -75,7 +75,8 @@ namespace DCGO.CardEffects.BT10
                             selectCardCoroutine: null),
                         },
                         remainingCardsPlace: RemainingCardsPlace.DeckBottom,
-                        activateClass: activateClass
+                        activateClass: activateClass,
+                        mutualConditions: true
                     ));
                 }
             }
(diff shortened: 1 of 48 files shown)


x89rt2 (approved):

x89rt2:
Old list I had:

BT3-008 Zubamon
BT3-062 Ludomon
BT3-093 Davis Motomiya
BT5-009 Shoutmon
BT8-058 Agumon
BT9-090 Maki Himekawa
BT16-039 Pulsemon
BT18-007 Gazimon
BT18-030 Candlemon
BT19-017 Sangomon
EX2-047 ADR-03 Pendulum Feet
EX3-031 Veedramon
EX7-008 ToyAgumon
EX7-016 Bulucomon
EX8-057 DemiDevimon
RB1-005 Gammamon
RB1-011 Jellymon
RB1-020 Angoramon
ST10-04 Gatomon
ST10-12 LadyDevimon
ST18-04 Pteromon
ST19-03 Shoemon

PR #924, opened 2026-03-23
Update ContinuousController.cs

x89rt2:
What if I just....


hooperk:
Have you tested locally to see if these exist?
Nvm, I see the point is to allow the user to provide 3 different images (or us if we can find an artist)

hooperk (approved):

PR #931, opened 2026-03-23
Fix Desync Issue

hooperk:


x89rt2:
Assuming absolutely everything else with the build is good, I'd probably revert this, send to public, then remerge with this fix for next beta session.

hooperk:
> Assuming absolutely everything else with the build is good, I'd probably revert this, send to public, then remerge with this fix for next beta session.

Absolutely

hooperk:
@x89rt2 since we do have AD1 issues to merge and retest, may as well look at merging this in too

x89rt2 (approved):

PR #941, opened 2026-03-25
Fix BT24-090 & BT24-094 throne rooms performance

hooperk:
Create Alliance Static Effect

Tested locally with Abyss Sanctuary, saw no lag and effect worked. Could not see if Alliance correctly showed in right click window due to missing textures.


Assets/Scripts/Script/CardEffects/AllianceClass.cs
hooperk, line 1: The Alliance Class was already not used by any code that did work, so I have removed it. It needs to be an Activate Class anyway for ordering.

Assets/Scripts/Script/Permanent.cs
```diff
@@ -2885,21 +2885,22 @@ public bool HasAlliance
...
+            
             foreach (Player player in GManager.instance.turnStateMachine.gameContext.Players_ForTurnPlayer)
             {
                 foreach (Permanent permanent in player.GetFieldPermanents())
                 {
                     #region Permanent Effects
                     foreach (ICardEffect cardEffect in permanent.EffectList(EffectTiming.OnAllyAttack))
                     {
-                        if (cardEffect is IAllianceEffect)
+                        if (cardEffect.EffectName == "Alliance")
```
hooperk, line 2899: This was not working at all since long ago Alliance was changed to an Activate Class

x89rt2 (approved):

PR #951, opened 2026-03-26
Fix AD1_021, BT7_032 not use OPT and fix Twice Per Turn

### LINKED ISSUE #472: Conditional Effects with Once Per Turns can have timing issues or possibly just use up the OPT incorrectly.

hooperk:
Marcus was tested locally and confirmed working, others not tested yet

### DIFF
--- a/Assets/Scripts/Script/ICardEffect.cs
+++ b/Assets/Scripts/Script/ICardEffect.cs
@@ -1087,7 +1087,6 @@ public static class ActivateICardEffectExtensionClass
             ((ICardEffect)activateICardEffect).SetOnProcessCallbuck(null);
 
             //Handling Effect
-            ((ICardEffect)activateICardEffect).EffectSourceCard.cEntity_EffectController.RegisterUseEfffectThisTurn(((ICardEffect)activateICardEffect));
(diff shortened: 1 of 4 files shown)


x89rt2:
@hooperk What makes this one different to https://github.com/DCGO2/DCGO/pull/636 ? The location? On first look it just looks like it's outside something.

x89rt2 (approved):

Assets/Scripts/Script/ICardEffect.cs
```diff
@@ -1087,7 +1087,6 @@ public static IEnumerator Activate_Execute(this ActivateICardEffect activateICar
             ((ICardEffect)activateICardEffect).SetOnProcessCallbuck(null);
 
             //Handling Effect
-            ((ICardEffect)activateICardEffect).EffectSourceCard.cEntity_EffectController.RegisterUseEfffectThisTurn(((ICardEffect)activateICardEffect));
```
hooperk, line 1090: @x89rt2 This is the difference FYI. This line was registering the effect as used a second time per usage.

I have removed the bug of each usage counting as 2 usages, which is why Twice per Turn should also be fixed by this PR

PR #952, opened 2026-03-26
Implement BT24-004, When would Link, and Link cost Reduction

hooperk:
Includes an overhaul to how link costs are paid, including a new Cost Reduction, which then will not conflict with regular cost reduction.

As part of this overhaul, the 2 existing cards which link for a reduced cost have been updated.

DoGatchmon was required as it can overlap with the Tapmon, stacking the cost reduction.


Michael8818 (approved):
lgtm

PR #953, opened 2026-03-27
Create ST24_15.cs

### LINKED ISSUE #810: ST24-15 DNA Charge
Image

x89rt2:


Assets/Scripts/CardEffect/ST24/White/ST24_15.cs
```diff
@@ -0,0 +1,251 @@
...
+            string SharedEffectName = "Play 1 [DATA SQUAD] card with a play cost of 4 or less from hand or trash for free, place this card in battle area";
+
+            string SharedEffectDescription(string tag) => $"[{tag}] You may play 1 [DATA SQUAD] trait card with a play cost of 4 or less from your hand or trash without paying the cost. Then, place this card in the battle area.";
+
+            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
+            {
+                bool CanSelectPlayCondition(CardSource cardSource)
+                {
+                    return (cardSource.IsDigimon
+                            || cardSource.IsTamer)
```
hooperk: HasPlayCost is more correct now than checking for being a Digimon or Tamer with Dual Cards and who knows what in the future. This same effect on E-Pulse will be able to play the new Dual level 5 if it has the same code, which is a bug, and there could be a cheap enough dual Data Squad level 5 as well

Assets/Scripts/CardEffect/ST24/White/ST24_15.cs
```diff
@@ -0,0 +1,251 @@
...
+                    return CardEffectCommons.CanTriggerSecurityEffect(hashtable, card);
+                }
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    yield return ContinuousController.instance.StartCoroutine(SharedActivateCoroutine(hashtable, activateClass));
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.PlaceDelayOptionCards(card: card, cardEffect: activateClass));
+                }
+            }
+            #endregion
```
hooperk, line 150: No reason to write a shared effect and duplicate this, instead use CardEffectCommons.AddActivateMainOptionSecurityEffect and have the full main effect in one place

Assets/Scripts/CardEffect/ST24/White/ST24_15.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            #region Start of Your Main Phase
+            if (timing == EffectTiming.OnStartMainPhase)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Place this card under 1 [DATA SQUAD] tamer face down to <Draw 1> and gain 1 memory", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, true, EffectDescription());
```
hooperk: Can No select true and no OPT, so lets set isOptional false

hooperk (changes requested):

Assets/Scripts/CardEffect/ST24/White/ST24_15.cs
x89rt2: we need to decide what route we are going to do with this since now we have your new button.

hooperk: I'm intending to add an extra flag .SetIsSkippable, so we can keep using false here to save clicks, but still include them under the new button

Assets/Scripts/CardEffect/ST24/White/ST24_15.cs
```diff
@@ -0,0 +1,235 @@
...
+                    return CardEffectCommons.CanTriggerOptionMainEffect(hashtable, card);
+                }
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    bool CanSelectPlayCondition(CardSource cardSource)
+                    {
+                        return (cardSource.IsDigimon
+                                || cardSource.IsTamer)
+                            && cardSource.HasPlayCost
```
hooperk, line 52: We don't need to check IsDigimon and IsTamer if we are checking HasPlaycost. The card itself does not state any requirement for the card to be a Digimon or Tamer

hooperk (changes requested):

Assets/Scripts/CardEffect/ST24/White/ST24_15.cs
```diff
@@ -0,0 +1,235 @@
...
+                        yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+
+                        IEnumerator SelectPermanentCoroutine(Permanent permanent)
+                        {
+                            selectedPermanent = permanent;
+
+                            yield return null;
+                        }
+
+                        if (selectedPermanent != null && !selectedPermanent.IsToken)
```
hooperk, line 217: Not going to delay this PR as it does not matter at all for this optional effect, but per the CRM a token is not even a valid target to choose for effects that place under a digimon.

hooperk (approved):

PR #977, opened 2026-03-30
shuffling improvements

Llandy3d:
These changes should add around `18.4~ quintillions` shuffle possibilities compared to the current limitation of `1.5~ billions` and in addition it introduces rejection sampling that should help in making players feel less that they have the same cards and help with variety.
The main issues stemmed from unity random functions that work in 32 bits with the 128bit PRNG, the logic in this PR brings a custom class to elevate randomness to current standards. Fun fact it's still close to 0% of all the possibilities of a 50 card deck but it should feel a lot more varied :)

Summary

  - Replace weak `DateTime.Now.Ticks` seed with cryptographically secure 64-bit seed via `System.Security.Cryptography.RandomNumberGenerator`
  - Replace `UnityEngine.Random` (Xorshift128, 32-bit seed) with Xoshiro256** (256-bit state) for all game-critical randomness
  - Add rejection sampling to eliminate modulo bias in card index selection

  Problem

  The current shuffling system has three compounding weaknesses:

  1. Insufficient shuffle space — `UnityEngine.Random.InitState()` takes a 32-bit int, capping the PRNG at ~4.3 billion possible sequences. A 50-card deck has 50! ≈ 3×10⁶⁴ possible orderings. The
   vast majority of legal shuffles can never occur, which constrains the distribution and causes repeated patterns players can feel.

  2. Non-cryptographic PRNG — Unity's `UnityEngine.Random` uses Xorshift128, a fast but statistically weak generator. Its 128-bit internal state can be reconstructed from observed outputs, and it
   fails some standard randomness test suites (BigCrush, PractRand).

### DIFF
--- a/Assets/Scripts/Script/ContinuousController.cs
+++ b/Assets/Scripts/Script/ContinuousController.cs
@@ -4,6 +4,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Threading.Tasks;
 using UnityEngine;
 using UnityEngine.Events;
@@ -494,9 +495,9 @@ public class ContinuousController : MonoBehaviour
     public async void Init()
     {
         Application.targetFrameRate = 60;
-        int random = RandomUtility.getRamdom();
-        UnityEngine.Random.InitState(random);
-        Debug.Log($"Game Initialize - random number sequence initialization,InitState:{random}");
+        long random = RandomUtility.GetSecureRandom();
+        GameRandom.Seed(random);
+        Debug.Log($"Game Initialize - random number sequence initialization, GameRandom.Seed:{random}");
 
         Sprite reverseCardSprite = await StreamingAssetsUtility.GetSprite("card_back_main");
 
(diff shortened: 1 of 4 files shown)


Michael8818:
Can you explain a bit more into "rejection sampling to eliminate modulo bias" and what that means to the end user?

Llandy3d:
> Can you explain a bit more into "rejection sampling to eliminate modulo bias" and what that means to the end user?

sure, the rejection sampling part might be the less important change in the PR but it's still the mathematically correct way to generate a random integer from a random byte stream.
In short the PNRG would give you a random int from 0 to 4 billion for example, if you want a number from 0 to 2 you would modulo by 3 (random_int % 3), the value 0 is more likely to appear but this is over the course of the calculationg happening a lot of times (millions/billions).

So for normal users it should be negligible, if you would analyze data over the course of many games you would start to see the bias towards some results.
In short since it's a casual game you can live without it, but any competitive game or poker for example would implement this to remove the bias, and since the cost is low/non-existant I added it as well!

x89rt2 (approved):

PR #978, opened 2026-03-30
Implement BT25-104 and convenience methods

hooperk:
Planning a follow up to this where I try to timestamp at least Change DP Class for ordering it. Will do follow up as a second big PR, so restricting this to just the cards and needed methods

### DIFF
--- a/Assets/Scripts/CardEffect/BT25/Red/BT25_104.cs
+++ b/Assets/Scripts/CardEffect/BT25/Red/BT25_104.cs
@@ -117,7 +117,7 @@ namespace DCGO.CardEffects.BT25
 
             string SharedHashString = "BT25_104_WD_WA";
             
-            string SharedEffectName = "1 Enemy Digimon gets -15k DP, play 1 tamer.";
+            string SharedEffectName = "As an option effect, 1 Enemy Digimon gets -15k DP, play 1 tamer.";
 
             string SharedEffectDescription(string tag) => $"[{tag}] [Once Per Turn] Activate 1 [Main] effect on this card's Option side.";
 
--- a/Assets/Scripts/CardEffect/BT25/Red/BT25_104.cs
+++ b/Assets/Scripts/CardEffect/BT25/Red/BT25_104.cs
@@ -1,0 +1,296 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+// Shinegreymon: Burst Mode // Final Shining Burst
+namespace DCGO.CardEffects.BT25
+{
+    public class BT25_104 : CEntity_Effect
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Script/CardEffectCommons.cs
```diff
@@ -686,6 +686,28 @@ public static void AddActivateMainOptionSecurityEffect(CardSource card, ref List
 
     #endregion
 
+    #region Activate Main of Dual Cards Option Side
+    public static IEnumerator ActivateMainOfOptionSide(CardSource card, ICardEffect activateClass, Func<ICardEffect, IEnumerator> afterMainEffect = null, bool asEffectOfThisDigimon = false)
+    {
+        ActivateClass mainActivateClass = OptionMainEffect(card);
+
+        #region Set as option effect unless explicitly effect of the digimon
```
hooperk, line 711: I hope and I pray tbh

Assets/Scripts/Script/CardEffectFactory/ChangeOriginDP.cs
```diff
@@ -8,6 +8,32 @@ public partial class CardEffectFactory
...
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        return ChangeBaseDPGlobalEffect(PermanentCondition, changeValue, isInheritedEffect, card, CanUseCondition);
+    }
+    #endregion
```
hooperk, line 33: This code that shows as added is actually the same as before. The rest of the code of this class was moved down to the other class, which then shows me changing what these methods were to more general ones.
The net result should be that this method behaves the exact same, but now there is another method for 1 card changing the baseDP of several others.

Assets/Scripts/CardEffect/BT25/Red/BT25_104.cs
```diff
@@ -0,0 +1,295 @@
...
+            {
+                cardEffects.Add(CardEffectFactory.BarrierSelfEffect(false, card, null));
+            }
+            #endregion
+
+            #region Shared WD / WA
+
+            string SharedHashString = "BT25_104_WD_WA";
+            
+            string SharedEffectName = "1 Enemy Digimon gets -15k DP, play 1 tamer.";
```
x89rt2: If you are going to do this description, it might be worth adding something like "as an option effect".

hooperk:
```suggestion
            string SharedEffectName = "As an option effect, 1 Enemy Digimon gets -15k DP, play 1 tamer.";
```

x89rt2 (approved):

PR #984, opened 2026-03-31
Improve CCU efficiency & various networking quality of life updates & fixes

Llandy3d:
The goal of this PR was to improve the efficiency of networking usage to allow more players but I ended up adding more improvements and fixing more issues for a better experience.

- Connection to proton are only made when actually searching for a game, otherwise the user gets disconnected everywhere else saving slots for more users. (now if you played a game and left the app in the menu, you would be consuming 1 CCU)
- Fixed an issue where the player would attempt to join a full room and be stuck forever, now it rejoins the lobby and creates a new room
- Fixed an issue where when max CCU was reached, the player would be stuck searching forever. Now it retries with a 3 second falloff for 5 times. A message is shown when retrying
- In details disconnection now happens when:
  - User navigates to home page
  - User cancels matchmaking
  - User ended a game
- F1 mapping to debug information on the top left of the screen for troubleshooting networking issues
- Debug listener for proton to log events

### DIFF
--- a/Assets/Scripts/Script/ContinuousController.cs
+++ b/Assets/Scripts/Script/ContinuousController.cs
@@ -1225,6 +1225,10 @@ public class ContinuousController : MonoBehaviour
             Application.Quit();
         }
 #endif
+        if (Input.GetKeyDown(KeyCode.F1))
+        {
+            _showPhotonDebug = !_showPhotonDebug;
+        }
     }
     int _frameCount = 0;
     int _updateFrame = 40;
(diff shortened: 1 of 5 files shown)


Llandy3d:
Please do test that room functionality is working correctly

x89rt2:
I'm seeing a conflict.

Llandy3d:
> I'm seeing a conflict.

thanks, I've incorporated the banlist change in my code

Assets/Scripts/Script/ContinuousController.cs
```diff
@@ -1334,7 +1338,137 @@ IEnumerator SetRandomCoroutine(int random)
...
+            Debug.Log("[Photon Debug] Created Room");
+        }
+        public void OnLeftRoom()
+        {
+            Debug.Log("[Photon Debug] Left Room");
+        }
+        public void OnFriendListUpdate(List<Photon.Realtime.FriendInfo> friendList) { }
+    }
+
+    void OnGUI()
```
Michael8818, line 1425: This is a pretty old method for UI within unity. It will become obsolete after some version updates planned to happen at some point. I think it is fine to keep for now, but reworking would be really nice.

Assets/Scripts/Script/ContinuousController.cs
```diff
@@ -1261,6 +1261,10 @@ private void Update()
             Application.Quit();
         }
 #endif
+        if (Input.GetKeyDown(KeyCode.F1))
+        {
+            _showPhotonDebug = !_showPhotonDebug;
+        }
```
Michael8818, line 1268: put this in a preprocessor directive such as below:

#if UNITY_EDITOR

#endif

Assets/Scripts/Script/ContinuousController.cs
```diff
@@ -1334,7 +1338,137 @@ IEnumerator SetRandomCoroutine(int random)
         Debug.Log($"random number sequence initialization,InitState:{random}");
     }
 
+    #region Photon Debug HUD
```
Michael8818, line 1344: if we can put this whole region in a preprocessor directive I think it goes a long way to make sure this isn't accessible outside of the editor (yes this should be added in MANY other places too) such as below:

#if UNITY_EDITOR

#endif

Michael8818 (changes requested):

Assets/Scripts/Script/ContinuousController.cs
Llandy3d: yeah I had the doubt between allowing this command in finals builds or just development builds, but limiting it to the editor should be fine as well!

Assets/Scripts/Script/ContinuousController.cs
Llandy3d: I tried to research when it's getting deprecated but seems that unity is just encouraging to use the other toolkit while they still use these methods internally for the editor, since it's for a simple in-editor debug message now I'm opting to leaving it as is and it shouldn't impact at all a version upgrade 🤞

Michael8818 (approved):
LGTM

PR #986, opened 2026-03-31
Small fix on use req cards.

x89rt2:
I'm seeing the definition already has a check that it's your own stuff so I'm removing them here, plus I suspect the cardeffectcommons might be an issue.

also fix soymp going face up instead of face down on ST24-15

cardeffectcommons was not the final issue, it was a card entity data error that has since been fixed

### DIFF
--- a/Assets/Scripts/CardEffect/P/Purple/P_238.cs
+++ b/Assets/Scripts/CardEffect/P/Purple/P_238.cs
@@ -11,16 +11,15 @@ namespace DCGO.CardEffects.P
         {
             List<ICardEffect> cardEffects = new List<ICardEffect>();
 
-            #region Ignore Color Requirment
+            #region Ignore Color Requirement
             if (timing == EffectTiming.None)
             {
                 cardEffects.Add(CardEffectFactory.UseRequirements(card, PermanentCondition));
 
                 bool PermanentCondition(Permanent permanent)
                 {
-                    return CardEffectCommons.IsOwnerPermanent(permanent, card)
-                        && permanent.TopCard.HasCSTraits
-                        && (permanent.IsDigimon || permanent.IsTamer);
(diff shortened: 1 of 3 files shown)


x89rt2:
Actually, verify for me but I don't think we even need the is tamer || digimon check. The fact that userequirements has 

CardEffectCommons.HasMatchConditionOwnersPermanent(card, permanentCondition)
                || CardEffectCommons.HasMatchConditionOwnersBreedingPermanent(card, permanentCondition);
                
means it's not counting placed options, no? If not, why is it not set up this way?

hooperk:
@x89rt2 If you don't mind I am going to make a change to make Use Req match RAW (and create a central place so if they do confirm not to count options later it will be an easy change for us)
The change I am planning will make this PR redundant

x89rt2:
> @x89rt2 If you don't mind I am going to make a change to make Use Req match RAW (and create a central place so if they do confirm not to count options later it will be an easy change for us) The change I am planning will make this PR redundant

I do have a fix on ST24-15 of

                            yield return ContinuousController.instance.StartCoroutine(selectedPermanent.AddDigivolutionCardsBottom(new List<CardSource>() { card }, activateClass, false, true));
                            
but sure, go for it.

hooperk:
> > @x89rt2 If you don't mind I am going to make a change to make Use Req match RAW (and create a central place so if they do confirm not to count options later it will be an easy change for us) The change I am planning will make this PR redundant
> 
> I do have a fix on ST24-15 of
> 
> ```
>                         yield return ContinuousController.instance.StartCoroutine(selectedPermanent.AddDigivolutionCardsBottom(new List<CardSource>() { card }, activateClass, false, true));
> ```
> 
> but sure, go for it.

I see, I'll capture the change to ST24_15 in my PR as well

PR #991, opened 2026-04-01
Fix link cards cannot link

hooperk:
Confirmed with local testing


x89rt2:
What was the reasoning behind this breaking?

hooperk:
> What was the reasoning behind this breaking?

I accidentally never assigned the link card, instead assigning a field that was currently null to itself, so when you got down to the part that actually tries to attach the card, it was attaching Nothing to the permanent

x89rt2 (approved):

PR #1001, opened 2026-04-02
BT18-095 Bug Fix Suggestion

LewisWelch94:


Assets/Scripts/CardEffect/BT18/Red/BT18_095.cs
```diff
@@ -110,130 +81,133 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
+                                    yield return null;
                                 }
 
-                                else
+                                bool SelectHybridCardConditionCustom(CardSource cardSource)
+                                {
+                                    return SelectHybridCardCondition(cardSource) &&
+                                           !selectedCards.Exists(selected =>
+                                               selected.CardNames.Exists(name =>
+                                                   cardSource.CardNames.Contains(name)));
```
hooperk, line 151: Use cardSource.EqualsCardName, CardNames.Contains causes confusion

hooperk: I also think to avoid issues if there is ever a valid digimon with multiple names we would need to do selectedCards.All(selected => selected.CardNames.Any(name => !cardSource.EqualsCardName(name)))

Assets/Scripts/CardEffect/BT18/Red/BT18_095.cs
```diff
@@ -110,130 +81,133 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
-                            IEnumerator SelectCardCoroutine(List<CardSource> cardSources)
+                            if (canSelectHand || canSelectTrash)
                             {
-                                digivolutionCards.AddRange(cardSources);
+                                if (canSelectHand && canSelectTrash)
+                                {
+                                    List<SelectionElement<bool>> selectionElements = new List<SelectionElement<bool>>()
+                                    {
+                                        new SelectionElement<bool>(message: $"From hand", value : true, spriteIndex: 0),
+                                        new SelectionElement<bool>(message: $"From trash", value : false, spriteIndex: 1),
```
hooperk, line 114: Needs third option to stop placing for this up to effect

Assets/Scripts/CardEffect/BT18/Red/BT18_095.cs
```diff
@@ -110,130 +81,133 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
                                 }
 
-                                if (CardEffectCommons.IsPermanentExistsOnBattleArea(tamerPermanent))
-                                {
-                                    yield return ContinuousController.instance.StartCoroutine(tamerPermanent
-                                        .AddDigivolutionCardsBottom(digivolutionCardsFixed, activateClass));
-                                }
+                                #endregion
+
+                                if (selectedCards.Count >= 1) yield return ContinuousController.instance.StartCoroutine(tamerPermanent
```
hooperk, line 209: We should have an additional select to order the cards afterwards

hooperk (changes requested):

Assets/Scripts/CardEffect/BT18/Red/BT18_095.cs
LewisWelch94: Ordering is done in the first select, select message does mention it.

Assets/Scripts/CardEffect/BT18/Red/BT18_095.cs
LewisWelch94: Hmm, that should be doable via the select itself, as this part is just for selecting if your doing it from hand or trash

Assets/Scripts/CardEffect/BT18/Red/BT18_095.cs
hooperk: Ah, your two responses have highlighted another issue: You should be allowed take from any combination of hand and trash, but this is only allowing you to take from one

hooperk: We have some examples of effects allowed to take from multiple zones, EX11-073 ExMaquinamon is the one I can remember easiest

PR #1010, opened 2026-04-03
fix double timer speed when searching for game

Llandy3d:


Assets/Scripts/Script/LobbyManager_RandomMatch.cs
```diff
@@ -151,7 +150,6 @@ IEnumerator TimeCountUp()
 
         while (!DoneCompleteMatching)
         {
-            time += (int)Time.deltaTime;
```
Llandy3d, line 154: dead code most of the time, this should be reporting always zero except in case of extreme lag, the real increment is done at the end of the function

Assets/Scripts/Script/LobbyManager_RandomMatch.cs
```diff
@@ -121,7 +121,6 @@ public IEnumerator Init()
         ReturnButton.SetActive(true);
         MessageText.text = "";
         TimeText.gameObject.SetActive(true);
-        StartCoroutine(TimeCountUp());
```
Llandy3d, line 124: This is a duplicated timer that was making `time` increment twice a second

x89rt2 (approved):

x89rt2:
@Llandy3d Connecting is always at 20 seconds and matching starts at 0 but increments by 2.

PR #1036, opened 2026-04-05
Implement BT25-097

### LINKED ISSUE #950: BT25-097 Guardian Palace
Image

hooperk:
Scapegoat presents a problem. Alliance could be reworked to a static method of the security card because it only needs to trigger once. Scapegoat needs to be able to separately trigger for every valid TS Digimon deleted simultaneously, and be ordered as needed. This means Scapegoat needs to be a granted effect on the Digimon.

Added a way to limit how often Add Skill class and its associated CanUse is called when the granted effect is relevant for only a single timing

### DIFF
--- a/Assets/Scripts/CardEffect/BT25/Yellow/BT25_097.cs
+++ b/Assets/Scripts/CardEffect/BT25/Yellow/BT25_097.cs
@@ -67,7 +67,7 @@ namespace DCGO.CardEffects.BT25
 
                 AddSkillClass addSkillClass = new AddSkillClass();
                 addSkillClass.SetUpICardEffect("Your Digimon gain Scapegoat", CanUseCondition, card);
-                addSkillClass.SetUpAddSkillClass(cardSourceCondition: CardSourceCondition, getEffects: GetEffects);
+                addSkillClass.SetUpAddSkillClass(cardSourceCondition: CardSourceCondition, getEffects: GetEffects, limitTiming: EffectTiming.WhenPermanentWouldBeDeleted);
                 cardEffects.Add(addSkillClass);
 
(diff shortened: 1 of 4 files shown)


Assets/Scripts/CardEffect/BT25/Yellow/BT25_097.cs
```diff
@@ -0,0 +1,302 @@
...
+
+                bool HasOXII(Permanent permanent)
+                {
+                    return permanent.TopCard.ContainsCardName("Junomon");
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistInSecurity(card, false)
+                        && CardEffectCommons.HasMatchConditionOwnersPermanent(card, HasOXII);
```
x89rt2, line 81: I feel like this is on if juno is in breeding? Although the CardSourceCondition also confuses me so idk.

hooperk: HasMatchConditionOwnersPermanent only sees battle area

x89rt2 (approved):

x89rt2:
I'll let you merge.

Assets/Scripts/CardEffect/BT25/Yellow/BT25_097.cs
hooperk: CardSourceCondition is an unfortunate side effect of AddSkillClass applying to cards and not permanents, but basically it is applying only to cards in the battle area, belonging to the owner of Guardian Palace, if the card is the top card of the permanent, and that permanent matches the Permanent condition of being a Purple or Yellow TS digimon

PR #1062, opened 2026-04-11
Update P_214.cs

x89rt2:
Fix tucks under breeding.

### DIFF
--- a/Assets/Scripts/CardEffect/P/Blue/P_214.cs
+++ b/Assets/Scripts/CardEffect/P/Blue/P_214.cs
@@ -64,7 +64,7 @@ namespace DCGO.CardEffects.P
 
             bool CanTuckTargetCondition(Permanent permanent)
             {
-                return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
+                return IsPermanentExistOnOwnerBattleAreaDigimon(permanent,card)
                     && CanSelectOwnSeadramonInName(permanent)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/P/Blue/P_214.cs
```diff
@@ -64,8 +64,9 @@ bool SharedCanActivateCondition(Hashtable hashtable)
 
             bool CanTuckTargetCondition(Permanent permanent)
             {
-                return CanSelectOwnSeadramonInName(permanent) &&
-                    permanent != card.PermanentOfThisCard();
+                return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
```
hooperk:
```suggestion
                return IsPermanentExistOnOwnerBattleAreaDigimon(permanent,card)
```

hooperk (changes requested):

Assets/Scripts/CardEffect/P/Blue/P_214.cs
x89rt2: Card directly mentions digimon so I'd like to have that. Should I just do your change and then add && isdigimon?

hooperk: Sure, check Digimon, but currently you're checking the wrong entity entirely

hooperk (approved):

PR #1063, opened 2026-04-11
Fix LKM Ace not granting Collision, Collision not displayed

hooperk:
Also fix Collision giving breeding area blocker

### DIFF
--- a/Assets/Scripts/CardEffect/BT18/Purple/BT18_083.cs
+++ b/Assets/Scripts/CardEffect/BT18/Purple/BT18_083.cs
@@ -155,7 +155,7 @@ namespace DCGO.CardEffects.BT18
 
             #region All Turns
 
-            if (timing == EffectTiming.None)
+            if (timing == EffectTiming.OnCounterTiming)
             {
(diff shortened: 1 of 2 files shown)


Assets/Scripts/CardEffect/BT18/Purple/BT18_083.cs
```diff
@@ -155,7 +155,7 @@ IEnumerator SelectCardCoroutine(CardSource cardSource)
 
             #region All Turns
 
-            if (timing == EffectTiming.None)
+            if (timing == EffectTiming.OnCounterTiming)
```
x89rt2, line 158: Can you run this one by me? Cause taking it at face value doesn't seem correct.

hooperk: Collision exists at this timing, rather than at the EffectTiming.None. I could have changed it, but didn't want to alter that many cards - every card that had Collision would have had to have had the timing changed to EffectTiming.None if I made it care about being there, and any I missed would not work just as LKM currently does not work. It was easier to continue keeping it at the same timing as when it was an activate ability. It doesn't actually do anything at that timing, but the game knows to check for it there

x89rt2 (approved):

PR #1069, opened 2026-04-12
Update BT25_097.cs

x89rt2:


hooperk:
What is this intending to do? It appears to just be adding an unused method

x89rt2:
> What is this intending to do? It appears to just be adding an unused method

The method was unused. It was a 50/50 on whether to call it or to delete it. I guess I should delete it instead?

Ok yea, it's play for free. Delete it it is.

hooperk (approved):

PR #1182, opened 2026-04-14
Add script for creating issues for an entire set

hooperk:
Simple script to create issues for an entire set once we know how many cards it has. I didn't see if there was a specific folder for scripts we run, so left it in root for now to make it easy to find and use

### DIFF
--- a/create_set.sh
+++ b/create_set.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+#Prerequisites: need to download github cli and will need bash
+#On windows `Git for Windows` from git-scm comes with bash
+#Must be logged in with gh auth login
+#./create_set.sh ST25 15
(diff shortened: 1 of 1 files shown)


x89rt2:
I had an issue with mass uploading issues that got me suspended for a bit. Did you verify this script follows github upload rate terms of service?

hooperk:
> I had an issue with mass uploading issues that got me suspended for a bit. Did you verify this script follows github upload rate terms of service?

Hmm, I wasn't aware of those, But I did add all of EX11 using it. Maybe after adding that whole set it doesn't like more

x89rt2:
@hooperk Have you kept using this script? Any issues with github on too many requests in a short time or did it have a delay to not run into the issue? Can this be merged?

PR #1223, opened 2026-04-19
Create BT25_024.cs

### LINKED ISSUE #1087: BT25-024 Lekismon
Image

x89rt2:


Assets/Scripts/CardEffect/BT25/Blue/BT25_024.cs
```diff
@@ -0,0 +1,117 @@
...
+                    cardEffect: activateClass).Draw()
+                );
+            }
+            #endregion
+
+            #region Your Turn
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Place 1 [Mineral]/[Rock] trait card from hand or trash under played/digivolved digimon", CanUseCondition, card);
```
hooperk: Wrong effect name

Assets/Scripts/CardEffect/BT25/Blue/BT25_024.cs
```diff
@@ -0,0 +1,117 @@
...
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CanSelectCardCondition);
+                }
+
+                bool PermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
+                        && permanent.TopCard.CardColors.Contains(CardColor.Red);
```
hooperk: Color is only checked at activate time (if, not when)

hooperk (changes requested):

Assets/Scripts/CardEffect/BT25/Blue/BT25_024.cs
x89rt2: Since the effect is fully optional, I don't want to waste people's time if the played one isn't the right color. Can you think of a scenario where this would matter?

hooperk: Yes, digivolving into a digimon and then being de-digivolved during that chain into a valid digimon, since de-digivolve would not trigger this effect again

x89rt2: Something like this? There might be a more eloquent way of doing it.... but I probably can't figure it out in a short time frame.

hooperk (approved):

PR #1253, opened 2026-04-23
Create BT25_014.cs

### LINKED ISSUE #956: BT25-014 Meramon
Image

x89rt2:


Assets/Scripts/CardEffect/BT25/Red/BT25_014.cs
```diff
@@ -0,0 +1,197 @@
...
+
+                string EffectDescription()
+                {
+                    return "[When Attacking] [Once Per Turn] Delete 1 of your opponent's Digimon with 4000 DP or less.";
+                }
+
+                bool CanSelectPermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card)
+                        && permanent.DP <= card.Owner.MaxDP_DeleteEffect(4000, activateClass);
```
hooperk, line 157: permanent.HasDP

hooperk (changes requested):

Assets/Scripts/CardEffect/BT25/Red/BT25_014.cs
x89rt2: Hm... I think some older cards might not have that, but eh, sure, I'll add it.

hooperk: It probably shouldn't matter since all Digimon should have DP (unless something weird happens) but the issue is if something does not have DP then permanent.DP returns -1, which is in fact less than the other number

hooperk (approved):

PR #1255, opened 2026-04-23
Fix AD1-005 can delete equal DP

hooperk:


x89rt2:
Already bypassed rules merged.

PR #1284, opened 2026-04-27
Implement BT25-077 and improvements to sharedeffectfactory

### LINKED ISSUE #1018: BT25-077 Bacchusmon
Image

hooperk:
Realized somewhat overdue that I should enable the sharedeffectfactory to pass through ActivateClass so we can still do things like check CanPlayPermanent in additionalUse/ActivateConditions

### DIFF
--- a/Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
+++ b/Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
@@ -91,7 +91,7 @@ namespace DCGO.CardEffects.BT25
 
             #region Shared OP / WD
 
-            string SharedEffectName = "You may play 1 [TS] Dgiimon with 6k or less DP";
+            string SharedEffectName = "You may play 1 [TS] Digimon with 6k or less DP";
 
             string SharedEffectDescription(string tag)
                 => $"[{tag}] You may play 1 [TS] trait Digimon card with 6000 DP or less from your hand without paying the cost.";
--- a/Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
+++ b/Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
@@ -150,7 +150,7 @@ namespace DCGO.CardEffects.BT25
                 activateClass.SetUpICardEffect("May suspend 1 Digimon. By Effect: Delete lowest DP enemy digimon", CanUseCondition, card);
                 activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, false, EffectDescription());
                 activateClass.SetIsSkippable(true);
-                activateClass.SetHashString("AD1_024_AT");
+                activateClass.SetHashString("BT25_077_AT");
                 cardEffects.Add(activateClass);
 
                 string EffectDescription() 
--- a/Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
+++ b/Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
@@ -1,0 +1,234 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System;
+
+// Bacchusmon
+namespace DCGO.CardEffects.BT25
+{
+    public class BT25_077 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Alt Digivolution
+            if (timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent permanent)
+                {
+                    return permanent.TopCard.HasTSTraits;
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(PermanentCondition, 3, true, card, null, level: 5));
+            }
+            #endregion
+
+            #region Reduce Play Cost
+            if (timing == EffectTiming.None)
+            {
+                ChangeCostClass changeCostClass = new ChangeCostClass();
+                changeCostClass.SetUpICardEffect("Play Cost -5", CanUseCondition, card);
+                changeCostClass.SetUpChangeCostClass(changeCostFunc: ChangeCost, cardSourceCondition: CardSourceCondition, rootCondition: RootCondition, isUpDown: isUpDown, isCheckAvailability: () => false, isChangePayingCost: () => true);
+                changeCostClass.SetNotShowUI(true);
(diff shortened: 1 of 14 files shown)


Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+
+            CardEffectFactory.ActivateClassesForSharedEffects
+                (ref cardEffects, timing, card,
+                    SharedEffectName,
+                    SharedActivateCoroutine,
+                    SharedEffectDescription,
+                    additionalActivateCondition: AdditionalActivateCondition,
+                    optional: false,
+                    onPlay: true,
```
x89rt2, line 143:
```suggestion
                    optional: false,
                    isSkippable: true,
                    onPlay: true,
```

Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
```diff
@@ -0,0 +1,234 @@
...
+            #endregion
+
+            #region All Turns
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("May suspend 1 Digimon. By Effect: Delete lowest DP enemy digimon", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, false, EffectDescription());
+                activateClass.SetIsSkippable(true);
+                activateClass.SetHashString("BT25_077_AT");
```
x89rt2, line 154: Do you have isskippable smart enough that if it's played by effect it won't be skippable since the deletion is mandatory?

Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
```diff
@@ -0,0 +1,234 @@
...
+                    if (isByEffect && CardEffectCommons.HasMatchConditionOpponentsPermanent(card, IsOpponentsWeakestDigimon))
+                    {
+                        SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+
+                        selectPermanentEffect.SetUp(
+                            selectPlayer: card.Owner,
+                            canTargetCondition: IsOpponentsWeakestDigimon,
+                            canTargetCondition_ByPreSelecetedList: null,
+                            canEndSelectCondition: null,
+                            maxCount: 1,
```
x89rt2:
```suggestion
                    if (isByEffect)
                    {
                        SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
                        
                        int maxCount = Math.Min(1, CardEffectCommons.HasMatchConditionPermanentCount(card, IsOpponentsWeakestDigimon));

                        selectPermanentEffect.SetUp(
                            selectPlayer: card.Owner,
                            canTargetCondition: IsOpponentsWeakestDigimon,
                            canTargetCondition_ByPreSelecetedList: null,
                            canEndSelectCondition: null,
                            maxCount: maxCount ,
```

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
hooperk: Is skippable no... but this is exactly why I didn't remove the isOptionalOverride function :D

Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
hooperk: Not quite right, it'll soft lock, but you are correct that we need to set executed regardless

Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
x89rt2: So, just to make sure cause I know very little about isskippable, should it be on this one or not? cause my impression is no

hooperk: It should not... and hang on I think I added stuff in the wrong place there

hooperk: A merge messed up the code a little

x89rt2: this optionaloverride is new to me, run it by me.

hooperk: It's that thing I tried before we got OPT refresh working. Where we were making it optional if it only had optional parts or not optional if it had any mandatory parts

x89rt2: Why are we using this method again? Why on this and not on marcus?

hooperk: Hmm... I could actually do one better and make it still not query the user like an optional: true effect

hooperk: OK. The method is now SetIsSkippableFunction.
It allows you to use a function to set whether or not the effect can be skipped so that it actually checks at effect selection time. In this example we say it is skippable as long as it was not triggered by effect

Assets/Scripts/CardEffect/ST23/Yellow/ST23_05.cs
```diff
@@ -37,15 +37,12 @@ bool PermanentCondition(Permanent targetPermanent)
...
                                                               SharedEffectDescription,
                                                               false,
-                                                              isOptionalFunction: IsOptional,
                                                               maxCountPerTurn: 1,
                                                               hashValue: SharedHashString,
                                                               whenDigivolving: true,
                                                               whenAttacking: true
                                                               );
 
-            bool IsOptional(Hashtable hashtable) => !CardEffectCommons.HasMatchConditionPermanent(IsWeakestEnemyDigimon);
```
hooperk, line 47: This was actually entirely wrong before. The effect will always attempt to remove an enemy digimon even if they don't have one, ex[pending the OPT.
I've checked the code below checks there is a valid target before attempting to do anything to avoid softlocks

Assets/Scripts/CardEffect/BT25/Black/BT25_077.cs
hooperk: > Why are we using this method again? Why on this and not on marcus?

Only just saw this one... we absolutely could and maybe should start adding this now to Marcus and others

x89rt2 (approved):

PR #1291, opened 2026-04-29
BT25-019

### LINKED ISSUE #957: BT25-019 UltimateBrachiomon
Image

LewisWelch94:


Assets/Scripts/Script/PermanentEffectFactory.cs
```diff
@@ -79,6 +76,35 @@ bool SkillCondition(ICardEffect cardEffect)
...
+        bool CardCondition(CardSource cardSource)
+        {
+            return cardSource == permanent.TopCard
+                && CardEffectCommons.IsExistOnBattleAreaDigimon(permanent.TopCard);
+        }
+
+        bool SkillCondition(ICardEffect cardEffect)
+        {
+            return CardEffectCommons.IsOpponentEffect(cardEffect, permanent.TopCard)
+                && cardEffect.IsOptionEffect;
```
hooperk: IsOptionEffect is only actually set on a handful of Three Musketeers cards to get them to work properly under digimon, and I'm not sure that even worked right. You will actually have to do !cardEffect.IsDigimonEffect && !cardEffect.IsTamerEffect

Assets/Scripts/Script/CardSource.cs
```diff
@@ -4143,6 +4143,18 @@ public bool HasBeatBreakTraits
...
 
     #endregion
+
+    #region whether this card has Dinosaur trait
+
+    public bool HasDinosaurTraits
+    {
+        get
+        {
+            return EqualsTraits("Dinosaur");
```
hooperk: I actually don't think we should do this one. After the whole thing where having Dragon traits was any Digimon with "Dragon" or "Saur" or "Cerotopsian" we should reserve this for if there is ever something along the lines of Containing "Saur" or Cerotopsian" but not "Dragon"

Assets/Scripts/CardEffect/BT25/Red/BT25_019.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            #endregion
+
+            #region OP/WD Shared
+            string SharedEffectName = "Delete 1 Digimon with highest DP.";
+            string SharedEffectDescription(string tag) => $"[{tag}] Delete 1 of your opponent's highest DP Digimon.";
+
+            bool CanSelectPermamentCondition(Permanent permanent)
+                => CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card)
+                    && CardEffectCommons.IsMaxDP(permanent, card.Owner.Enemy, null);
```
hooperk: IsMaxDP checks IsPermanentExistsOnOpponentBattleAreaDigimon so don't need both

Assets/Scripts/CardEffect/BT25/Red/BT25_019.cs
```diff
@@ -0,0 +1,134 @@
...
+                onPlay: true,
+                whenDigivolving: true);
+
+            #endregion
+
+            #region End of Your Turn
+            if (timing == EffectTiming.OnEndTurn)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("If 5 or more memory, this gains immune to digimon effects, if 5 or less memory, this gains immune to option effects.", CanUseCondition, card);
```
x89rt2: This description makes it sound like it's checking your side of the mem gauge when it's checking opponent's

x89rt2 (changes requested):

Assets/Scripts/Script/PermanentEffectFactory.cs
x89rt2: @hooperk we good here?

Assets/Scripts/Script/CardSource.cs
x89rt2: @hooperk we good here?

Assets/Scripts/Script/PermanentEffectFactory.cs
hooperk: Yeah, it was already changed to check that is is not a digimon or tamer effect, which is how we check for is option

Assets/Scripts/Script/CardSource.cs
hooperk: Yup, this code is removed. You can see "Outdated" on the comment when the code is no longer the same

x89rt2 (approved):

hooperk (approved):

PR #1316, opened 2026-05-02
Update BT25_074.cs

x89rt2:
Attempt to fix cost reduction soft lock on WD/WA and added missing AT.

### DIFF
--- a/Assets/Scripts/CardEffect/BT25/Black/BT25_074.cs
+++ b/Assets/Scripts/CardEffect/BT25/Black/BT25_074.cs
@@ -121,7 +121,7 @@ namespace DCGO.CardEffects.BT25
 
                 bool CardSourceCondition(CardSource cardSource)
                 {
-                    return cardSource == card;
+                    return true;
                 }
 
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT25/Black/BT25_074.cs
```diff
@@ -121,11 +121,7 @@ bool PermanentsCondition(List<Permanent> targetPermanents)
 
                 bool CardSourceCondition(CardSource cardSource)
                 {
-                    return cardSource.IsDigimon
-                        && cardSource.HasPlayCost
-                        && cardSource.GetCostItself <= 12
-                        && (cardSource.EqualsTraits("D-Brigade")
-                            || cardSource.EqualsTraits("ACCEL"));
+                    return cardSource == card;
```
hooperk: This was correct-ish before. Now it would apply when playing the Tankdramon itself. Actually, we already know whatever card is safe because we selected it, so we don't need to double check conditions agian.
```suggestion
                    return true;
```

hooperk (changes requested):

PR #1331, opened 2026-05-03
Update move condition check to activate condition, reduce clicks

hooperk:
Technically the cards work correctly, this is just to improve UX

### DIFF
--- a/Assets/Scripts/CardEffect/BT25/Blue/BT25_024.cs
+++ b/Assets/Scripts/CardEffect/BT25/Blue/BT25_024.cs
@@ -1,5 +1,6 @@
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 // Lekismon
 namespace DCGO.CardEffects.BT25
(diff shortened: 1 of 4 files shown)


Assets/Scripts/CardEffect/BT25/Blue/BT25_024.cs
```diff
@@ -70,7 +72,20 @@ bool CanUseCondition(Hashtable hashtable)
                 bool CanActivateCondition(Hashtable hashtable)
                 {
                     return CardEffectCommons.IsExistOnBattleArea(card)
-                        && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CanSelectCardCondition);
+                        && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CanSelectCardCondition)
+                        && HasValidColor(hashtable);
```
hooperk, line 76: Problem before was checking them as part of trigger condition. It is safe to check as part of activate condition

x89rt2 (approved):

PR #1355, opened 2026-05-05
BT25-077 Bug Fix

LewisWelch94:


hooperk:
Was this causing a problem? I would expect that using CardDP is correct since that is the one that takes into account effects that change the DP of cards in hand, but when there aren't any of those it starts from BaseCardDP so should have the same value

LewisWelch94:
> Was this causing a problem? I would expect that using CardDP is correct since that is the one that takes into account effects that change the DP of cards in hand, but when there aren't any of those it starts from BaseCardDP so should have the same value

From looking at the method, it was trying to find the card on the battle area, as we use it alot with permanent.topcard.
it also has a base value of -1 in the method, so my guess is its no finding it, using math.min which is 0, and check is getting set to <= 0 dp due to it.

Changing it from CardDP to BaseCardDP instantly fixed this, as was able to summon out multiple ts mons under 6k using heroms, both with the op & wd

hooperk:
> > Was this causing a problem? I would expect that using CardDP is correct since that is the one that takes into account effects that change the DP of cards in hand, but when there aren't any of those it starts from BaseCardDP so should have the same value
> 
> From looking at the method, it was trying to find the card on the battle area, as we use it alot with permanent.topcard. it also has a base value of -1 in the method, so my guess is its no finding it, using math.min which is 0, and check is getting set to <= 0 dp due to it.
> 
> Changing it from CardDP to BaseCardDP instantly fixed this, as was able to summon out multiple ts mons under 6k using heroms, both with the op & wd
Even if that were the case, and I don't believe it is as nothing is checking for the card to be in play, 0 is already less than 6000 so it would have been allowing all cards

LewisWelch94:
> > > Was this causing a problem? I would expect that using CardDP is correct since that is the one that takes into account effects that change the DP of cards in hand, but when there aren't any of those it starts from BaseCardDP so should have the same value
> > 
> > 
> > From looking at the method, it was trying to find the card on the battle area, as we use it alot with permanent.topcard. it also has a base value of -1 in the method, so my guess is its no finding it, using math.min which is 0, and check is getting set to <= 0 dp due to it.
> > Changing it from CardDP to BaseCardDP instantly fixed this, as was able to summon out multiple ts mons under 6k using heroms, both with the op & wd
> > Even if that were the case, and I don't believe it is as nothing is checking for the card to be in play, 0 is already less than 6000 so it would have been allowing all cards

Thats fair, but something is going wrong with that previous method either way.

hooperk:
Marking as Draft so Xeight doesn't just merge when they see it later

PR #1362, opened 2026-05-06
Update BT25-075

hooperk:


Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
```diff
@@ -282,6 +282,7 @@ bool PermanentCondition(Permanent permanent)
                 activateClass.SetUpICardEffect("Linked digimon may Attack", CanUseCondition, card);
```
x89rt2, line 282: Can we not put the card name here as well or is that not possible with current code? I know this wouldn't be perfect if there are 2 cards with same name, but every little helps.

hooperk: That is what line 308 is doing.

x89rt2 (approved):

PR #1367, opened 2026-05-06
Fix BT20-043 & BT20-036 cost reduction

hooperk:


Assets/Scripts/CardEffect/BT20/Yellow/BT20_036.cs
```diff
@@ -58,15 +58,7 @@ string EffectDiscription()
 
                 bool CardCondition(CardSource cardSource)
                 {
-                    if (cardSource == card)
-                    {
-                        if (CardEffectCommons.IsExistOnHand(cardSource))
-                        {
-                            return CardEffectCommons.HasMatchConditionPermanent(HasAccelTraitInPlay);
```
hooperk, line 65: FYI this is checked in the CanActivateCondition, and the change cost function, does not need to be checked here

x89rt2 (approved):

PR #1379, opened 2026-05-08
Hide Attack option when card has {Breeding}[Main]

hooperk:


x89rt2:
@hooperk If we added something like if !can attack, would it also remove the option from bt11 vemmon if he was stunned?

x89rt2 (approved):

PR #1388, opened 2026-05-08
Update EX7_048.cs

x89rt2:
Modernize and possibly fix sending used beelstar dual to top or bot.

### DIFF
--- a/Assets/Scripts/CardEffect/EX7/Black/EX7_048.cs
+++ b/Assets/Scripts/CardEffect/EX7/Black/EX7_048.cs
@@ -2,6 +2,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 
+// Gundramon
 namespace DCGO.CardEffects.EX7
 {
     public class EX7_048 : CEntity_Effect
@@ -11,12 +12,11 @@ namespace DCGO.CardEffects.EX7
             List<ICardEffect> cardEffects = new List<ICardEffect>();
 
             #region Alternate Digivolution
-
             if (timing == EffectTiming.None)
             {
                 bool PermanentCondition(Permanent targetPermanent)
                 {
-                    return targetPermanent.TopCard.IsLevel5 && targetPermanent.TopCard.HasText("Three Musketeers");
+                    return targetPermanent.TopCard.HasText("Three Musketeers");
                 }
 
                 cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/EX7/Black/EX7_048.cs
```diff
@@ -51,123 +61,38 @@ bool CanSelectCardSharedCondition(CardSource cardSource)
...
-                    }
+                    yield return ContinuousController.instance.StartCoroutine(
+                        CardEffectCommons.PlayOptionCards(
+                            cardSources: new List<CardSource>() { cardSource },
+                            activateClass: activateClass,
+                            payCost: false,
+                            root: SelectCardEffect.Root.Library
+                        )
+                    );
+                    yield return null;
```
hooperk, line 93: I can't really tell when everything was removed, what was the change to fix the current issue?

hooperk, line 93: Also minor and just weirdness of IEnumerators, but an IEnumerator only needs to have 1 `yield return` anywhere within it, although more does not hurt. We only need `yield return null;` when we don't yield return something anywhere else in it.

x89rt2: Old Code

yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.SimplifiedRevealDeckTopCardsAndSelect(
    revealCount: 6,
    simplifiedSelectCardConditions:
    new SimplifiedSelectCardConditionClass[]
    {
        new(
            canTargetCondition: CanSelectCardSharedCondition,
            message: "Select 1 Option card with the [Three Musketeers] trait.",
            mode: SelectCardEffect.Mode.Custom,
            maxCount: 1,
            selectCardCoroutine: SelectCardCoroutine),
    },
    remainingCardsPlace: RemainingCardsPlace.DeckTopOrBottom,
    activateClass: activateClass,
    canNoSelect: true
    
    New code
    
    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.SimplifiedRevealDeckTopCardsAndSelect(
                    revealCount: 6,
                    simplifiedSelectCardConditions:
                    new SimplifiedSelectCardConditionClass[]
                    {
                        new SimplifiedSelectCardConditionClass( <----- I saw this code in other simplifiedreveal cards
                            canTargetCondition: CanSelectCardSharedCondition,
                            message: "Select 1 Option card with the [Three Musketeers] trait.",
                            mode: SelectCardEffect.Mode.Custom,
                            maxCount: 1,
                            selectCardCoroutine: SelectCardCoroutine),
                    },
                    remainingCardsPlace: RemainingCardsPlace.DeckTopOrBottom,
                    activateClass: activateClass,
                    canNoSelect: true

x89rt2: This is mostly copied code from the old one. If you think something can be removed, feel free to remove it yourself so I can see it and try to understand it's logic.

hooperk: Ah. Yeah this is 1:1 what was there before, C# does not require us to declare the full class name when using new. Still the clean up is nice and will help when we can dig in further

hooperk (approved):

PR #1389, opened 2026-05-09
Fix playing multiple digimon and digixros/assembly

hooperk:
Been able to test that single Digixros still works correctly, Assembly should be the exact same then.
Need a deck for testing playing multiple at once.

Works off the principle that no effect that plays multiple digimon at once makes you pay for them, so in those scenarios it is safe to move the digixros/assembly to after when costs are paid, in this case immediately before playing each individual card.

### DIFF
--- a/Assets/Scripts/Script/CardController.cs
+++ b/Assets/Scripts/Script/CardController.cs
@@ -738,23 +738,28 @@ public class PlayCardClass
 
             #endregion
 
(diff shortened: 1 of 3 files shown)


x89rt2:
Has issues: https://discord.com/channels/839041812932329512/1445092870934499329/1502659474736021695

x89rt2 (approved):

PR #1393, opened 2026-05-09
Fix retaliation allow ties, checks is in trash

hooperk:


Assets/Scripts/Script/CardEffectCommons/KeyWordEffects/Retaliation.cs
```diff
@@ -21,7 +21,7 @@ public static bool CanActivateRetaliation(Hashtable hashtable)
 
                     if (TopCard != null)
                     {
-                        if (IsByBattle(hashtable))
```
hooperk, line 24: IsByBattle just gets the BattleHashtable and checks it isn't null. As it happens that is what the next 2 lines of code do anyway, so just removed that. Separately, this was the best spot to check that the top card is still in trash

x89rt2 (approved):

PR #1404, opened 2026-05-10
Fix BT25-042

hooperk:


Assets/Scripts/CardEffect/BT25/Yellow/BT25_042.cs
```diff
@@ -109,7 +109,7 @@ IEnumerator SuccessProcess(List<CardSource> cardSources)
             if (timing == EffectTiming.OnLoseSecurity)
             {
                 ActivateClass activateClass = new ActivateClass();
-                activateClass.SetUpICardEffect("Play 1 level 4 or lower [Angel]/[Iliad] trait card from hand without cost, then 2 digimon gain reboot & blocker until opponent turn end", CanUseCondition, card);
+                activateClass.SetUpICardEffect("Play 1 level 4 or lower [Angel]/[Illad] trait card from hand without cost, then 2 digimon gain reboot & blocker until opponent turn end", CanUseCondition, card);
                 activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, false, EffectDescription());
```
hooperk, line 113: Latter half of the effect is always mandatory

Assets/Scripts/CardEffect/BT25/Yellow/BT25_042.cs
```diff
@@ -121,24 +121,20 @@ string EffectDescription()
...
-                        && cardSource.HasLevel
-                        && cardSource.Level <= 4
-                        && (cardSource.EqualsTraits("Angel")
-                            || cardSource.EqualsTraits("Iliad"))
-                        && CardEffectCommons.CanPlayAsNewPermanent(card, false, activateClass);
-                }
+                    => cardSource.HasLevel
+                       && cardSource.Level <= 4
+                       && (cardSource.EqualsTraits("Angel") || cardSource.HasIliadTraits)
+                       && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
```
hooperk, line 137: Actually issue was card (checking legality of playing ClavisAngemon) vs cardSource (check legality of playing the other card). It wasn't working when hard played because a card that is the only card in a stack cannot be removed to play to a new stack.

FYI, CanPlayAsNewPermanent is a rather expensive thing to check, we should check it last as much as possible. No point checking it for an Ogremon in hand

x89rt2 (approved):

PR #1407, opened 2026-05-10
Update BT14-029 reset OPT when not used and allow to whiff intentionally

hooperk:


Assets/Scripts/CardEffect/BT14/Blue/BT14_029.cs
```diff
@@ -42,69 +42,76 @@ bool CanSelectCardCondition(CardSource cardSource)
...
                             isFromOnly1Permanent: false,
                             activateClass: activateClass
                         ));
+                    }
                 }
             }
 
             if (timing == EffectTiming.OnAllyAttack)
             {
+                int digivolutionCardCount = card.PermanentOfThisCard().DigivolutionCards.Count;
```
hooperk: No. this will capture the amount when it is triggered, which could change by activation

x89rt2: so 2 of these 1 in coroutine and 1 in skippable function?

hooperk: Even better, SkippableFunction checks the same thing so just put it in the one place

x89rt2 (approved):

PR #1418, opened 2026-05-12
Create LM_066.cs

x89rt2:


Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+                        selectPlayer: card.Owner,
+                        canTargetCondition: CanSelectSuspendCondition,
+                        canTargetCondition_ByPreSelecetedList: null,
+                        canEndSelectCondition: null,
+                        maxCount: 1,
+                        canNoSelect: true,
+                        canEndNotMax: false,
+                        selectPermanentCoroutine: SelectPermanentCoroutine,
+                        afterSelectPermanentCoroutine: null,
+                        mode: SelectPermanentEffect.Mode.Tap,
```
hooperk: We should do this with Mode.Custom and SuspendPeremanentAndProcess... method

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+                        cardEffect: activateClass);
+
+                    IEnumerator SelectPermanentCoroutine(Permanent permanent)
+                    {
+                        selectedPermament = permanent;
```
hooperk, line 216: Used = true in here. Just attempting to suspend is using the effect, even if unsuccessful.

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+
+                string EffectDescription()
+                {
+                    return "When this card would be used, by suspending 2 Digimon, reduce the cost by 2.";
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerWhenPermanentWouldPlay(hashtable, cardSource => cardSource == card)
+                        && CardEffectCommons.MatchConditionPermanentCount(CanSelectSuspendCondition) >= 2;
```
hooperk: This line should be an activate condition. It is possible some other immediate effect could read "When you would use an option, unsuspend 1 digimon" which would make this payable

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+                ChangeCostClass changeCostClass = new ChangeCostClass();
+                changeCostClass.SetUpICardEffect("Use Cost -2", CanUseCondition, card);
+                changeCostClass.SetUpChangeCostClass(changeCostFunc: ChangeCost, cardSourceCondition: CardSourceCondition, rootCondition: RootCondition, isUpDown: isUpDown, isCheckAvailability: () => true, isChangePayingCost: () => true);
+                changeCostClass.SetNotShowUI(true);
+                cardEffects.Add(changeCostClass);
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnHand(card)
+                        && CardEffectCommons.MatchConditionPermanentCount(CanSelectSuspendCondition) >= 2;
```
hooperk, line 386: It is fine here, different to the activate class

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+                }
+
+                int ChangeCost(CardSource cardSource, int Cost, SelectCardEffect.Root root, List<Permanent> targetPermanents)
+                {
+                    if (CardSourceCondition(cardSource)
+                    && RootCondition(root))
```
hooperk: Both of these effects require a permanentCondition, or this will show as applying to digivolving

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+                            }
+
+                            return Cost;
+                        }
+
+                        bool PermanentsCondition(List<Permanent> targetPermanents)
+                        {
+                            return targetPermanents == null
+                                    || targetPermanents.Count((targetPermanent) => targetPermanent != null) == 0;
+                        }
```
hooperk, line 360: Copy this Permanent Condition to the one below

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,458 @@
...
+                            yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.GainCantUnsuspendUntilOpponentTurnEnd(
+                                targetPermanent: permanent,
+                                activateClass: activateClass
+                            ));
+                        }
+                    }
+
+                    int maxCount = Math.Min(CardEffectCommons.MatchConditionPermanentCount(CanSelectBotDeckCondition), Mathf.FloorToInt(CardEffectCommons.MatchConditionPermanentCount(CanCountDigimonCondition) / 2));
+
+                    if (maxCount > 0)
```
hooperk, line 495: nice

hooperk (changes requested):

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,477 @@
...
+
+                    selectPermanentEffect.SetUpCustomMessage("Select 1 Digimon to battle.", "The opponent is selecting 1 Digimon to battle.");
+
+                    IEnumerator SelectPermanentCoroutine(Permanent permanent)
+                    {
+                        yield return ContinuousController.instance.StartCoroutine(new IBattle(card.PermanentOfThisCard(), permanent, null, true).Battle());
+                    }
+
+                    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+                }
```
hooperk, line 152: This method needs a RemoveUse if you do neither

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,477 @@
...
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerWhenPermanentWouldPlay(hashtable, cardSource => cardSource == card);
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.MatchConditionPermanentCount(CanSelectSuspendCondition) >= 2;
```
hooperk, line 270: So apparently you are allowed choose targets that will fail due to immunity, so this condition needs to not include permanent.CanSuspend.

Awkwardly, it will need the CanSuspend if the cost reduction is needed to be a valid play

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,477 @@
...
+                        selectPlayer: card.Owner,
+                        canTargetCondition: CanSelectSuspendCondition,
+                        canTargetCondition_ByPreSelecetedList: null,
+                        canEndSelectCondition: null,
+                        maxCount: 2,
+                        canNoSelect: true,
+                        canEndNotMax: false,
+                        selectPermanentCoroutine: SelectPermanentCoroutine,
+                        afterSelectPermanentCoroutine: null,
+                        mode: SelectPermanentEffect.Mode.Tap,
```
hooperk: Will need Mode.Custom and SuspendPeremanentAndProcess.
IEnumerator SuccessProcess(List<Permanent> suspendedPermanents
And will need to check suspendedPermanents.Count == 2

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,477 @@
...
+
+                    SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+
+                    selectPermanentEffect.SetUp(
+                        selectPlayer: card.Owner,
+                        canTargetCondition: CanSelectSuspendCondition,
+                        canTargetCondition_ByPreSelecetedList: null,
+                        canEndSelectCondition: null,
+                        maxCount: 2,
+                        canNoSelect: true,
```
hooperk: Will need to depend on if the option can be afforded, need CanNoSelect() similar to Ghoulmon, which will also be needed for the PermanentCondition

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,477 @@
...
+
+                    IEnumerator SelectPermanentCoroutine(Permanent permanent)
+                    {
+                        selectedPermanents.Add(permanent);
+                        yield return null;
+                    }
+
+                    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+
+                    if (selectedPermanents.Count == 2 && selectedPermanents.All(permanent => permanent.IsSuspended))
```
hooperk: Ok... didn't see you already had this, we should still change in case something else is what suspends the cards somehow

hooperk (changes requested):

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
x89rt2: Unless you can think of a realistic situation where this matters, I'd rather keep it this way to save wasted time. I'll wait on further changes till your response.

hooperk: choose 2 targets to suspend 1 when you only have one that will succeed, and while you could afford the full cost

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,531 @@
...
+                    {
+                        canNoSelect = false;
+                    }
+
+                    bool CanTargetCondition_ByPreSelecetedList(List<Permanent> permanents, Permanent permanent)
+                    {
+                        foreach (Permanent permanent1 in permanents)
+                        {
+                            if (canNoSelect = false
+                            && !permanent1.CanSuspend)
```
hooperk, line 294: Nice

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,531 @@
...
+                        card.Owner.MaxMemoryCost)
+                    {
+                        canNoSelect = false;
+                    }
+
+                    bool CanTargetCondition_ByPreSelecetedList(List<Permanent> permanents, Permanent permanent)
+                    {
+                        foreach (Permanent permanent1 in permanents)
+                        {
+                            if (canNoSelect = false
```
hooperk:
```suggestion
                            if (canNoSelect == false
```

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
```diff
@@ -0,0 +1,531 @@
...
+                            activateClass,
+                            SuccessProcess,
+                            null));
+                    }
+
+                    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+
+                    IEnumerator SuccessProcess(List<Permanent> selectedPermanents)
+                    {
+                        if (card.Owner.CanReduceCost(null, card)) ContinuousController.instance.PlaySE(GManager.instance.GetComponent<Effects>().BuffSE);
```
hooperk, line 333: Need to check if selectedPermanents.Count == 2 to ensure both got suspended, not just 1

hooperk (changes requested):

Assets/Scripts/CardEffect/LM/Green/LM_066.cs
x89rt2: SuspendPeremanentAndProcessAccordingToResult doesn't verify that it fully completed? It gives the go ahead on partials too?

PR #1423, opened 2026-05-13
Update BT15_102.cs

x89rt2:
Modernizing and fixing some bugs.

Fixed only grabbing digimon.
Fixed comparing card IDs instead of names maybe.
Will attempt to fix only grabbing from trash and not battle area. (baseline has been done, permanent selection needs work and currently has errors.)

Will not be fixing any of the other listed bugs with this PR.

### DIFF
--- a/Assets/Scripts/CardEffect/BT15/White/BT15_102.cs
+++ b/Assets/Scripts/CardEffect/BT15/White/BT15_102.cs
@@ -1,3 +1,4 @@
+using DCGO.CardEntities;
 using System;
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT15/White/BT15_102.cs
```diff
@@ -311,82 +287,48 @@ bool isUpDown()
...
-                            {
-                                return true;
-                            }
-                        }
-                    }
+                    ICardEffect activateClass = card.EffectList(EffectTiming.BeforePayCost).Find(cardEffect => cardEffect.EffectName == "Placing 1 [Dark Masters] to get Play Cost -4");
 
-                    return false;
+                    return activateClass != null
+                        && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CanSelectCardCondition);
```
x89rt2: Does this not need a check battle area digimon match condition thing?

hooperk: Yes

Assets/Scripts/CardEffect/BT15/White/BT15_102.cs
```diff
@@ -350,6 +350,7 @@ bool isUpDown()
                 ActivateClass activateClass = new ActivateClass();
                 activateClass.SetUpICardEffect("Place 1 digimon from trash under this Digimon's digivolution cards", CanUseCondition, card);
```
hooperk: While we're looking at this card could we include activating the [On Play] as part of the description?

hooperk (changes requested):

Assets/Scripts/CardEffect/BT15/White/BT15_102.cs
x89rt2: I'll do it, I just hope I remember correctly that all of the floodgates look at the start of the description and no where else.

hooperk: Can you remember which digimon floodgate this and we could check?

x89rt2: valdur arm and aegisdramon

hooperk: Oh those. Yeah we made them check only the start of the description, they never checked effect name

x89rt2: Alright. Approve and merge if looks good to you then unless you want to take a shot at some of apoca's other bugs.

hooperk (approved):

PR #1432, opened 2026-05-15
Hybrid lvl 5s

x89rt2:
Wild guess at fixing the lvl 5 hybrids going back to hand if fail evo.

If method is correct, doing the other ones should be easy.

### DIFF
--- a/Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
+++ b/Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
@@ -362,12 +362,12 @@ namespace DCGO.CardEffects.BT12
                                                 payCost: true,
                                                 targetPermanent: selectedPermanent,
                                                 isTapped: false,
-                                                failedProcess: OnFail(),
+                                                failedProcess: OnFail(card),
                                                 root: SelectCardEffect.Root.Hand,
-                                                activateETB: true).PlayCard());              
+                                                activateETB: true).PlayCard());
                                         }
 
-                                        IEnumerator OnFail()
+                                        IEnumerator OnFail(CardSource cardSource)
                                         {
                                             yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddTrashCard(card));
                                         }
--- a/Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
+++ b/Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
@@ -362,7 +362,7 @@ namespace DCGO.CardEffects.BT12
                                                 payCost: true,
                                                 targetPermanent: selectedPermanent,
                                                 isTapped: false,
-                                                failedProcess: OnFail(card),
+                                                failedProcess: OnFail,
                                                 root: SelectCardEffect.Root.Hand,
                                                 activateETB: true).PlayCard());
                                         }
--- a/Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
+++ b/Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
@@ -362,8 +362,14 @@ namespace DCGO.CardEffects.BT12
                                                 payCost: true,
                                                 targetPermanent: selectedPermanent,
                                                 isTapped: false,
+                                                failedProcess: OnFail(),
                                                 root: SelectCardEffect.Root.Hand,
-                                                activateETB: true).PlayCard());
+                                                activateETB: true).PlayCard());              
+                                        }
+
+                                        IEnumerator OnFail()
+                                        {
+                                            yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddTrashCard(card));
                                         }
 
                                         #region 6
--- a/Assets/Scripts/Script/CardController.cs
+++ b/Assets/Scripts/Script/CardController.cs
@@ -967,7 +967,7 @@ public class PlayCardClass
 
                     if (_failedProcess != null)
                     {
-                        yield return ContinuousController.instance.StartCoroutine(_failedProcess);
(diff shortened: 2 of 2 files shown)


Assets/Scripts/Script/CardController.cs
```diff
@@ -954,6 +965,21 @@ bool IsShowEffect()
                         }
                     }
                 }
+
+                if(playFailed)
```
hooperk: There is an if (playFailed) immediately above this, we should put this code in that block and add the else after that block

Assets/Scripts/Script/CardController.cs
```diff
@@ -209,6 +218,8 @@ public void SetIsBreedingArea()
     int _burstTamerFrameID = -1;
     int[] _appFusionFrameIDs = null;
     bool _isBreedingArea = false;
+    IEnumerator _successProcess;
+    IEnumerator _failedProcess;
```
hooperk: Nice approach. Could we make these `IEnumerator<CardSource>` and pass the failed/succeeded card back in case we want to do anything with them? Especially for multiple simultaneous plays, it would allow us to do something for only the cards that fail, if something can make just 1 fail, and likewise for only the card that succeeds

hooperk (changes requested):

Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
```diff
@@ -362,8 +362,14 @@ bool PermanentCondition(Permanent permanent)
                                                 payCost: true,
                                                 targetPermanent: selectedPermanent,
                                                 isTapped: false,
+                                                failedProcess: OnFail(),
                                                 root: SelectCardEffect.Root.Hand,
-                                                activateETB: true).PlayCard());
+                                                activateETB: true).PlayCard());              
+                                        }
+
+                                        IEnumerator OnFail()
```
hooperk: To go with my suggestion, this will be IEnumerator OnFail(CardSource cardSource)

Assets/Scripts/Script/CardController.cs
x89rt2: Ill be honest, I have no idea how any of these work. I copied the melga warg ones as best I could and did some creative interpretations to get the edi to stop giving me errors.

x89rt2: But ill take another look after work.

hooperk: You did a good job. I believe this would work as is, but I'd like the way to pass back the attempted to play card to match some other functions

x89rt2: Like this?

Assets/Scripts/CardEffect/BT12/Red/BT12_015.cs
```diff
@@ -362,10 +362,16 @@ bool PermanentCondition(Permanent permanent)
                                                 payCost: true,
                                                 targetPermanent: selectedPermanent,
                                                 isTapped: false,
+                                                failedProcess: OnFail(card),
```
hooperk:
```suggestion
                                                failedProcess: OnFail,
```

Assets/Scripts/Script/CardController.cs
```diff
@@ -117,8 +117,15 @@ public IEnumerator Discard()
...
-    bool activateETB)
+    public PlayCardClass(List<CardSource> cardSources,
+        Hashtable hashtable,
+        bool payCost,
+        Permanent targetPermanent,
+        bool isTapped,
+        SelectCardEffect.Root root,
+        bool activateETB,
+        IEnumerator successProcess = null,
+        IEnumerator failedProcess = null)
```
hooperk:
```suggestion
        IEnumerator<CardSource> successProcess = null,
        IEnumerator<CardSource> failedProcess = null)
```

Assets/Scripts/Script/CardController.cs
```diff
@@ -131,6 +138,8 @@ public PlayCardClass(List<CardSource> cardSources, Hashtable hashtable, bool pay
         _isTapped = isTapped;
         Root = root;
         _activateETB = activateETB;
+        _successProcess = (IEnumerator<CardSource>)successProcess;
+        _failedProcess = (IEnumerator<CardSource>)failedProcess;
```
hooperk:
```suggestion
        _successProcess = successProcess;
        _failedProcess = failedProcess;
```

Assets/Scripts/Script/CardController.cs
```diff
@@ -953,6 +964,18 @@ bool IsShowEffect()
                             yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddTrashCard(cardSource));
                         }
                     }
+
+                    if (_failedProcess != null)
+                    {
+                        yield return ContinuousController.instance.StartCoroutine(_failedProcess);
```
hooperk:
```suggestion
                        yield return ContinuousController.instance.StartCoroutine(_failedProcess(cardSource));
```

Assets/Scripts/Script/CardController.cs
```diff
@@ -953,6 +964,18 @@ bool IsShowEffect()
...
+                    }
+                }            
+                else
+                {
+                    if (_successProcess != null)
+                    {
+                        yield return ContinuousController.instance.StartCoroutine(_successProcess);
```
hooperk:
```suggestion
                        yield return ContinuousController.instance.StartCoroutine(_successProcess(cardSource));
```

hooperk (changes requested):

x89rt2:
@hooperk 5 errors

hooperk (approved):

PR #1436, opened 2026-05-16
Add setisdigimoneffect to various cards

x89rt2:


hooperk (approved):
All of these should be getting set automatically, but no harm to set manually as well

PR #1458, opened 2026-05-21
Fix EX11-046 should delete when no highest play cost

hooperk:


x89rt2:
I had this one up https://github.com/DCGO2/DCGO/pull/1452

PR #1462, opened 2026-05-21
Fix compilation issues

hooperk:


Assets/Scripts/Script/CardController.cs
```diff
@@ -1053,6 +1054,11 @@ bool IsShowEffect()
 
         yield return ContinuousController.instance.StartCoroutine(useOption.UseOption());
 
+        if (_successProcess != null)
+        {
+            yield return ContinuousController.instance.StartCoroutine(_successProcess(optionCards));
+        }
```
hooperk, line 1060: Had to move these down here and make them act on the full list of whatever is actually played or used as otherwise it was going to do the success process before the card was actually played

x89rt2 (approved):

PR #1469, opened 2026-05-23
Create EX12_069.cs

### LINKED ISSUE #1165: EX12-069: Virus Buster
Image

x89rt2:
Issue with security your turn. Not sure how exactly to get it to do what I want.

### DIFF
--- a/Assets/Scripts/CardEffect/EX12/Yellow/EX12_069.cs
+++ b/Assets/Scripts/CardEffect/EX12/Yellow/EX12_069.cs
@@ -1,0 +1,177 @@
+using System.Collections;
+using System.Collections.Generic;
+
+// Virus Busters
+namespace DCGO.CardEffects.EX12
+{
+    public class EX12_069 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Use Req
+            if (timing == EffectTiming.None)
+            {
+                cardEffects.Add(CardEffectFactory.UseRequirements(card, CardCondition));
+
+                bool CardCondition(CardSource cardSource)
+                {
+                    return cardSource.EqualsTraits("VB");
+                }
+            }
+            #endregion
+
+            #region Your Turn Security
+            if (timing == EffectTiming.OnAllyAttack)
+            {
+                ActivateClass activateClass = new ActivateClass();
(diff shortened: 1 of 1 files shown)


x89rt2:
Maybe this second one's route is slightly better? I did fix the error, but doing so reminded me of the possible flaw of this method. I believe it captures the level early and if there is a when attacking digivolve, it will have the wrong level. The method is so that if there isn't a digimon in hand, not bothering asking for evo, but it's looking like it might not be possible, and it will need to activate always. Tell me if you get ideas.

Assets/Scripts/CardEffect/EX12/Yellow/EX12_069.cs
```diff
@@ -0,0 +1,166 @@
...
+
+                bool PermanentCondition(Permanent permanent)
+                    => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
+                    && permanent.TopCard.HasLevel
+                    && permanent.TopCard.Level >= 4
+                    && permanent.TopCard.EqualsTraits("VB");
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                    => cardSource.EqualsTraits("Shambala")
+                    && cardSource.Level == AttackingLevel;
```
hooperk: Need HasLevel as well, or else a Digimon with no level would be allowed play another digimon without a level

hooperk: Also CanPlayAsNewPermanent, with fixedCost: cardSource.GetCostItself - 3

hooperk (changes requested):

Assets/Scripts/CardEffect/EX12/Yellow/EX12_069.cs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            #endregion
+
+            #region Main Effect
+            if (timing == EffectTiming.OptionSkill)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Replace your bottom sec with this face-up card", CanUseCondition, card);
+                activateClass.SetUpActivateClass(null, ActivateCoroutine, -1, false, EffectDescription());
+                cardEffects.Add(activateClass);
```
hooperk, line 97: Doesn't matter, but FYI this entire effect could be replaced with CardEffectFactory.ReplaceBottomSecurityWithFaceUpOptionEffect(card)

hooperk (approved):

PR #1509, opened 2026-06-04
First attempt at solving issues with copy effects

hooperk:
An attempt at fixing an issue with copied affects, by giving them unique hashstrings so the OPTs behave correctly.
Also extracting to a factory method which can be improved over time, like if there's any weirdness with the conditions to use the granted keyword effects.
Lastly this method directly adds the effects to the card instead of using AddSkillEffect, to make the overall flow more efficient, as AddSkillEffect would be adding every single skill, not just the ones for the timing we are querying.


x89rt2 (approved):

sollybr:
https://discord.com/channels/1095717982522585098/1392508058411728937/1529382050874789939

PR #1554, opened 2026-06-22
Update EX12-004 with Execute

hooperk:


Assets/Scripts/Script/CardEffectFactory/KeyWordEffects/Execute.cs
```diff
@@ -46,7 +46,8 @@ public static ActivateClass ExecuteEffect(Permanent targetPermanent, bool isInhe
         bool CanUseCondition(Hashtable hashtable)
         {
             return CardEffectCommons.IsExistOnBattleArea(card) &&
-                   CardEffectCommons.IsOwnerTurn(card);
+                   CardEffectCommons.IsOwnerTurn(card) &&
+                   (condition == null || condition());
```
hooperk, line 50: This change is to make sure it only triggers if the digimon has execute at trigger time, as well as when it goes ot activate it

x89rt2 (approved):

PR #1560, opened 2026-07-03
Rework PlayForFree and PlayForCost modes into CEC method

hooperk:


x89rt2:
does this support if we want to add a aftercoroutine thing? like if not used reset opt?

x89rt2 (approved):

PR #1561, opened 2026-07-03
Update EX12-071

hooperk:


x89rt2:
Why does this matter?

hooperk:
> Why does this matter?

With return it was returning true or false based only on the first card in the hand, meaning if that card is not SW then no card could digivolve

x89rt2 (approved):

PR #1571, opened 2026-07-05
Update EX12-059 display effect

hooperk:


Assets/Scripts/Script/PermanentEffectFactory.cs
```diff
@@ -157,19 +158,26 @@ public static AddDetailClass AddDetailClass(Permanent targetPermanent, string de
     {
         if(targetPermanent == null || targetPermanent.TopCard == null) return null;
         
-        return CardEffectFactory.AddDetailClass(CanUseCondition, PermanentCondition, detail, triggerEffect, activateClass.EffectSourceCard);
```
x89rt2, line 160: There's like 4 cards that use CardEffectFactory.AddDetailClass, are we fine removing this?

hooperk: We're not removing it, this is just now going through the static version but it should be the same command. In fact literally the only thing the static endpoint does is call this same line

x89rt2 (approved):

PR #1576, opened 2026-07-07
Update EX11-025 can't place under Kongou

hooperk:


x89rt2:
hm.... doesn't PutSecurityBottom already have a check for that?

hooperk:
> hm.... doesn't PutSecurityBottom already have a check for that?

No it doesn't

x89rt2 (approved):

PR #1692, opened 2026-07-14
Update BT25_074.cs

sollybr:
Make [All turns] obligatory

Expected: Triggered and activated upon meeting conditions
image

Result: Trigger prompting user whether to activate
image


x89rt2:
I'm closing this one because I'm doing this fix along with another smaller fix to the card and it will be merged internally.

PR #1694, opened 2026-07-17
On Deletion ESS do not resolve if the card with the printed text gets Played to the battle area while the effect is pending.

### LINKED ISSUE #583: On Deletion ESS do not resolve, occurs when if the card gets played to the battle area while effect pending
Hand bounce is worked as intended

sollybr:
Property getter unexpectedly feeding incorrect hashtable to CanTrigger Causing RegisterOnDeletion to override correct value

CanActivate must consider that played ESS granting card must still trigger OD as long as Top Card remains in the trash

---

Timeline of debugging and error tracking with most relevant comment:
https://github.com/sollybr/DCGO/discussions/2#discussioncomment-17667852

---

Closes:  #583

### DIFF
--- a/Assets/Scripts/Script/ICardEffect.cs
+++ b/Assets/Scripts/Script/ICardEffect.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
@@ -391,7 +391,8 @@ public abstract class ICardEffect
                 {
                     if (IsInheritedEffect || IsLinkedEffect)
                     {
-                        if (EffectSourceCard == EffectSourceCard.PermanentOfThisCard().TopCard)
+                        bool CardBelowBecameTopMostCard = EffectSourceCard == EffectSourceCard.PermanentOfThisCard().TopCard;
+                        if (CardBelowBecameTopMostCard)
                             return false;
 
                         if (EffectSourceCard.IsFlipped)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/Script/ICardEffect.cs
```diff
@@ -391,7 +391,8 @@ public bool CanActivate(Hashtable hashtable)
                 {
                     if (IsInheritedEffect || IsLinkedEffect)
                     {
-                        if (EffectSourceCard == EffectSourceCard.PermanentOfThisCard().TopCard)
+                        bool CardBelowBecameTopMostCard = EffectSourceCard == EffectSourceCard.PermanentOfThisCard().TopCard;
```
hooperk, line 394: Potential NPE if the effect source card has no permanent, such as if it is the card in the trash

Assets/Scripts/Script/ICardEffect.cs
```diff
@@ -807,14 +802,12 @@ public bool IsOnDeletion
...
-                        Permanent effectPermanent = EffectSourceCard.PermanentOfThisCard() ?? new Permanent(new List<CardSource>() { EffectSourceCard });
+                        // Permanent effectPermanent = EffectSourceCard.PermanentOfThisCard() ?? new Permanent(new List<CardSource>() { EffectSourceCard });
 
-                        Hashtable hashtable = CardEffectCommons.OnDeletionHashtable(new List<Permanent>() { effectPermanent }, null, null, false);
+                        // Hashtable hashtable = CardEffectCommons.OnDeletionHashtable(new List<Permanent>() { effectPermanent }, null, null, false);
 
-                        if (CanTrigger(hashtable))
-                        {
+                        // if (CanTrigger(hashtable))
                             return true;
```
hooperk, line 810: This part is good, IsOnDeletion really does not need to be checking if it could trigger

hooperk (changes requested):

Assets/Scripts/Script/ICardEffect.cs
sollybr: That's a negative.
https://github.com/sollybr/DCGO/blob/03e1eb2961c439c4a152588ea754a46278136a88/Assets/Scripts/Script/ICardEffect.cs#L390
already checks for NPs in that context, as well as contexts immediately below.

Assets/Scripts/Script/ICardEffect.cs
sollybr: It's best that in the future, state checking methods don't arbitrarily get access to methods that modify later state checkers' targets.

hooperk (approved):

PR #1695, opened 2026-07-17
Vulcanusmon softlock

sollybr:
Give choice to player to select any available linked Digimon
---
What this fixes: unexpected exception softlocking game when there are no Digimon permanents
What this does not fix: OP/WD triggering YT twice (perhaps for each card linked), which caused the exception when Vulcanus was removed

### DIFF
--- a/Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
+++ b/Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
@@ -244,50 +244,52 @@ namespace DCGO.CardEffects.BT25
             if (timing == EffectTiming.WhenLinked)
             {
                 ActivateClass activateClass = new ActivateClass();
-                activateClass.SetUpICardEffect("Linked digimon may Attack", CanUseCondition, card);
+                activateClass.SetUpICardEffect("Linked Digimon may Attack", CanUseCondition, card);
                 activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDescription());
                 activateClass.SetIsSkippable(true);
                 activateClass.SetEffectTargets(TargetablePermanents);
+
                 cardEffects.Add(activateClass);
 
-                string EffectDescription()
-                {
-                    return "[Your Turn] When your Digimon get linked, one of them may attack.";
-                }
+                string EffectDescription() => "[Your Turn] When your Digimon get linked, one of them may attack.";
 
-                bool PermanentCondition(Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card);
+                bool CanSelectPermanentCondition(Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card);
 
-                bool CanUseCondition(Hashtable hashtable)
-                {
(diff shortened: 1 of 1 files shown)


sollybr:
Related problem is https://discord.com/channels/839041812932329512/1519046949326684374

Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
```diff
@@ -244,50 +244,52 @@ bool PermanentCondition(Permanent permanent)
...
-                }
+                bool CanSelectPermanentCondition(Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card);
 
-                Permanent GetAttacker(Hashtable hashtable) => CardEffectCommons.GetPermanentFromHashtable(hashtable);
+                bool CanUseCondition(Hashtable hashtable) => CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass) && 
+                        CardEffectCommons.IsOwnerTurn(card) && 
+                        CardEffectCommons.CanTriggerWhenLinked(hashtable, CanSelectPermanentCondition, null);
 
-                List<Permanent> TargetablePermanents(Hashtable hashtable) => new List<Permanent>() { GetAttacker(hashtable) };
+                List<Permanent> TargetablePermanents(Hashtable hashtable) => CardEffectCommons.GetPermanentsFromHashtable(hashtable).Where(permanent => permanent.CanAttack(activateClass)).ToList();
```
hooperk, line 262: This does not appear to be getting used, was the intention to use validTargets = TargetablePermanents(); validTargets.Contains instead of CanSelectPermanentCondition below?

hooperk (changes requested):

Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
sollybr: That should have been removed altogether. Probably escaped my attention when changing the mechanism and refactoring.

Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
```diff
@@ -247,7 +247,6 @@ bool PermanentCondition(Permanent permanent)
                 activateClass.SetUpICardEffect("Linked Digimon may Attack", CanUseCondition, card);
                 activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDescription());
                 activateClass.SetIsSkippable(true);
-                activateClass.SetEffectTargets(TargetablePermanents);
```
hooperk, line 250: Oh sorry, I missed that it was used here. This is actually the part that highlights which Digimon the given effect applies to, and until we rework the link trigger to handle multiple Digimon should remain so the user can tell what they are agreeing to attack with

hooperk (changes requested):

Assets/Scripts/CardEffect/BT25/Black/BT25_075.cs
sollybr: Oooh wait you are right. My lack of sleep mixed the meaning of "Effect" with "Skill".

hooperk (approved):

PR #1698, opened 2026-07-21
Create BT26_033

### LINKED ISSUE #1616: BT26-033: Jupitermon // Wide Plasment
Image

ThunderG99:
Resolves #1616

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
+++ b/Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
@@ -19,8 +19,7 @@ namespace DCGO.CardEffects.BT26
             {
                 bool PermanentCondition(Permanent targetPermanent)
                 {
-                    return targetPermanent.TopCard.IsLevel5
-                        && targetPermanent.TopCard.HasTSTraits;
+                    return targetPermanent.TopCard.HasTSTraits;
                 }
 
                 cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(
@@ -94,7 +93,6 @@ namespace DCGO.CardEffects.BT26
                 ActivateClass activateClass = new ActivateClass();
                 activateClass.SetUpICardEffect("Add top sec to hand, may play/use 1 [Iliad]/[TS] from hand cost -5", CanUseCondition, card);
                 activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDescription());
-                activateClass.SetHashString("BT26_033_WD");
                 cardEffects.Add(activateClass);
 
                 string EffectDescription()
@@ -104,13 +102,13 @@ namespace DCGO.CardEffects.BT26
 
                 bool CanUseCondition(Hashtable hashtable)
                 {
-                    return CardEffectCommons.CanTriggerWhenDigivolving(hashtable, card)
-                        && CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+                    return CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass)
+                        && CardEffectCommons.CanTriggerWhenDigivolving(hashtable, card);
                 }
 
                 bool CanActivateCondition(Hashtable hashtable)
                 {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+                    return CardEffectCommons.IsExistOnBattleAreaActivate(card, activateClass);
                 }
 
                 IEnumerator ActivateCoroutine(Hashtable hashtable)
@@ -132,7 +130,6 @@ namespace DCGO.CardEffects.BT26
                     #region If your turn, play/use Iliad cost -5
                     if (CardEffectCommons.IsOwnerTurn(card))
                     {
-                        bool isUsed = false;
                         CardSource selectedCard = null;
 
                         bool CanSelectCardCondition(CardSource cardSource)
@@ -176,8 +173,6 @@ namespace DCGO.CardEffects.BT26
 
                         if (selectedCard != null)
                         {
-                            isUsed = true;
-
                             #region Reduce Cost by 5
                             IEnumerator ReduceCost()
                             {
@@ -236,7 +231,6 @@ namespace DCGO.CardEffects.BT26
                             }
                         }
 
-                        if (!isUsed) activateClass.RemoveUse();
                     }
                     #endregion
                 }
@@ -248,8 +242,7 @@ namespace DCGO.CardEffects.BT26
             {
                 ActivateClass activateClass = new ActivateClass();
                 activateClass.SetUpICardEffect("By placing top stacked card as bottom security, card doesn't leave", CanUseCondition, card);
-                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDescription());
-                activateClass.SetHashString("BT26_033_AT_Protect_TS");
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, true, EffectDescription());
                 cardEffects.Add(activateClass);
 
                 string EffectDescription()
@@ -259,20 +252,21 @@ namespace DCGO.CardEffects.BT26
 
                 bool CanUseCondition(Hashtable hashtable)
                 {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
+                    return CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass)
                         && CardEffectCommons.CanTriggerWhenPermanentRemoveField(hashtable, PermanentCondition);
                 }
 
                 bool CanActivateCondition(Hashtable hashtable)
                 {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
-                        && card.PermanentOfThisCard().DigivolutionCards.Count > 0;
+                    return CardEffectCommons.IsExistOnBattleAreaActivate(card, activateClass)
+                        && card.PermanentOfThisCard().DigivolutionCards.Count > 0
+                        && card.Owner.CanAddSecurity(activateClass);
                 }
 
                 bool PermanentCondition(Permanent permanent)
                 {
                     return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card)
-                        && (permanent.TopCard.IsDigimon || permanent.TopCard.IsTamer)
+                        && (permanent.IsDigimon || permanent.IsTamer)
                         && permanent.TopCard.HasTSTraits;
                 }
 
@@ -284,25 +278,17 @@ namespace DCGO.CardEffects.BT26
                     Permanent thisPermanent = card.PermanentOfThisCard();
                     CardSource topStacked = card;
 
-                    if (card.Owner.CanAddSecurity(activateClass))
+                    yield return ContinuousController.instance.StartCoroutine(GManager.instance.GetComponent<Effects>().RemoveDigivolveRootEffect(topStacked, thisPermanent));
+
+                    yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddSecurityCard(topStacked, toTop: false, faceUp: false));
+
+                    foreach (Permanent permanent in protectedPermanents)
                     {
-                        yield return ContinuousController.instance.StartCoroutine(GManager.instance.GetComponent<Effects>().RemoveDigivolveRootEffect(topStacked, thisPermanent));
-
-                        if (topStacked.IsACE)
-                        {
-                            yield return ContinuousController.instance.StartCoroutine(new AceOverflowClass(new List<CardSource> { topStacked }).Overflow());
-                        }
-
-                        yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddSecurityCard(topStacked, toTop: false, faceUp: false));
-
-                        foreach (Permanent permanent in protectedPermanents)
-                        {
-                            permanent.willBeRemoveField = false;
-                            permanent.HideDeleteEffect();
-                            permanent.HideHandBounceEffect();
-                            permanent.HideDeckBounceEffect();
-                            permanent.HideWillRemoveFieldEffect();
-                        }
+                        permanent.willBeRemoveField = false;
+                        permanent.HideDeleteEffect();
+                        permanent.HideHandBounceEffect();
+                        permanent.HideDeckBounceEffect();
+                        permanent.HideWillRemoveFieldEffect();
                     }
                 }
             }
--- a/Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
+++ b/Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
@@ -1,0 +1,422 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+// Jupitermon // Wide Plasment
+namespace DCGO.CardEffects.BT26
+{
+    public class BT26_033 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Digimon Effects
+
+            #region Alt Digivolution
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Digimon Effects
+
+            #region Alt Digivolution
+            if (timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.IsLevel5
```
x89rt2: the is level 5 is no longer used here and was replaced by the level 5 below. remove this one

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerWhenDigivolving(hashtable, card)
+                        && CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
```
x89rt2: These have been changed to IsExistOnBattleAreaTrigger and IsExistOnBattleAreaActivate respectively. These check situations where the card enters the battle area, trigger, leave the battle and re-enter it before activating. There are also IsExistOnBattleAreaDigimonTrigger versions, but I personally don't like adding the Digimon part on these incase some future effect comes out makes us have to go back to these and remove the Digimon. Make sure to fix this issue on all other cards you submitted if they have em.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+                                    cardSources: new List<CardSource>() { selectedCard },
+                                    activateClass: activateClass,
+                                    payCost: true,
+                                    isTapped: false,
+                                    root: SelectCardEffect.Root.Hand,
+                                    activateETB: true));
+                            }
+                        }
+
+                        if (!isUsed) activateClass.RemoveUse();
```
x89rt2: So these and the isUsed are used in case the effect is Once Per Turn and it has optional aspects that if you deny all shouldn't use up the OPT. It's not necessary here.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+            }
+            #endregion
+
+            #region All Turns Protect TS
+            if (timing == EffectTiming.WhenRemoveField)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("By placing top stacked card as bottom security, card doesn't leave", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDescription());
+                activateClass.SetHashString("BT26_033_AT_Protect_TS");
```
x89rt2: This effect is not OPT, does not need a hashstring and the 1 before true should be a -1

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+            }
+            #endregion
+
+            #region When Digivolving
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Add top sec to hand, may play/use 1 [Iliad]/[TS] from hand cost -5", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDescription());
+                activateClass.SetHashString("BT26_033_WD");
```
x89rt2: don't need a hashstring if an effect is not OPT

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
+                        && card.PermanentOfThisCard().DigivolutionCards.Count > 0;
+                }
+
+                bool PermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card)
+                        && (permanent.TopCard.IsDigimon || permanent.TopCard.IsTamer)
```
x89rt2:
```suggestion
                        && (permanent.IsDigimon || permanent.IsTamer)
```

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
+                        && CardEffectCommons.CanTriggerWhenPermanentRemoveField(hashtable, PermanentCondition);
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
```
x89rt2: same thing as above on the trigger and activate

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    List<Permanent> protectedPermanents = CardEffectCommons.GetPermanentsFromHashtable(hashtable)
+                                .Filter(PermanentCondition);
+
+                    Permanent thisPermanent = card.PermanentOfThisCard();
+                    CardSource topStacked = card;
+
+                    if (card.Owner.CanAddSecurity(activateClass))
```
x89rt2: This is probably better on the canactivatecondition to save on time

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+                    {
+                        yield return ContinuousController.instance.StartCoroutine(GManager.instance.GetComponent<Effects>().RemoveDigivolveRootEffect(topStacked, thisPermanent));
+
+                        if (topStacked.IsACE)
+                        {
+                            yield return ContinuousController.instance.StartCoroutine(new AceOverflowClass(new List<CardSource> { topStacked }).Overflow());
+                        }
```
x89rt2: I don't think this could ever be possible and if it ever does become possible, we are in big trouble. Remove it.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+                    cardSourceCondition: CardSourceCondition,
+                    rootCondition: RootCondition,
+                    isUpDown: () => false,
+                    isCheckAvailability: () => false,
+                    isChangePayingCost: () => true);
+                cardEffects.Add(changeCostClass);
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return card.Owner.HandCards.Contains(card);
```
x89rt2, line 329: I'm not fully keeping up with this effect but this card is ALWAYS checking security, not just from hand so this is wrong.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,422 @@
...
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    List<Permanent> destroyTargets = card.Owner.Enemy.GetBattleAreaDigimons().Filter(LowestDPCondition);
+
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DeletePeremanentAndProcessAccordingToResult(
+                        targetPermanents: destroyTargets,
+                        activateClass: activateClass,
+                        successProcess: null,
+                        failureProcess: null));
```
x89rt2, line 389: I feel like there's probably a better one we could use instead of this but I'm blanking right now so it can stay if you can't find it.

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done. Didn't realized this in testing as you can't "use" it but yeah, no point being opt.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done, same as the other one you can't really do it more than once, hence why I didn't realize this, but no reason for it to be opt.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: So I was confused about this too, turns out I based how this effects puts the top card in security from Temple of beginnings, and it has this control, so I seem to have transfered here. It's not doing anythere there either since we have no aegiochus/jupiter ACE, but it's futureproofing, and yeah makes zero sense here.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: Done.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
ThunderG99: I left it as is, let me know if you think of a better way of doing this.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,408 @@
...
+
+                bool RootCondition(SelectCardEffect.Root root)
+                {
+                    return true;
+                }
+
+                bool PermanentsCondition(List<Permanent> targetPermanents)
+                {
+                    return true;
+                }
```
hooperk, line 354:
```suggestion
                bool PermanentsCondition(List<Permanent> targetPermanents)
        {
            return targetPermanents == null || targetPermanents.Count(targetPermanent => targetPermanent != null) == 0;
        }
```
Otherwise this will apply to digivolution

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,408 @@
...
+            }
+            #endregion
+
+            #region Cancel security cost for digivolution only
+            // VERY HACKY PATCH: So this sucks right, Option Use Cost uses ChangeCostClass which affects ALL cost calculations for this card, including digivolution.
+            // This results in + security_count being added to digivolution costs too, which is not what I want (it should only affect the option part of the card play cost).
+            // I don't think there's a way to increase only the option part play cost, I think this is the only card that does anything like this.
+            // To somehow equalize the evo cost, we negate the security count specifically for digivolution costs using ChangeDigivolutionCostStaticEffect.
+            // This only affects digivolution, NOT option use .
+            // so the option cost correctly ends up as 2 + security_count while digivolution costs what it should (5 for red yellow 4 for TS).
```
hooperk, line 41: See below comment, we have a way to not apply these to digivolutions

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
hooperk:
```suggestion
                    return true;
```
This is still showing as checking hand for me

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
```diff
@@ -0,0 +1,408 @@
...
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    List<Permanent> destroyTargets = card.Owner.Enemy.GetBattleAreaDigimons().Filter(LowestDPCondition);
+
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DeletePeremanentAndProcessAccordingToResult(
+                        targetPermanents: destroyTargets,
+                        activateClass: activateClass,
+                        successProcess: null,
+                        failureProcess: null));
```
hooperk, line 389:
```suggestion
                    yield return ContinuousController.instance.StartCoroutine(destroyTargets, CardEffectCommons.CardEffectHashtable(activateClass)).Destroy());
```
When we don't care about the result we just call the delete directly.
This one is an optional change, tbh calling delete and process with null for both process still achieves the goal, and avoids accidentally passing wrong hashtable. This is more of an FYI while we decide if this would be the better way to recommend people go for this

hooperk (changes requested):

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
alyscs: Fixed via ThunderG99/DCGO#1 (opened against this branch since I don't have write access to it directly): PermanentsCondition now excludes non-null targetPermanents so this only applies to option use, not digivolution. As a result the 'Cancel security cost for digivolution only' hacky patch is now redundant and was removed.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
alyscs: Confirmed — same fix, see reply above.

Assets/Scripts/CardEffect/BT26/Yellow/BT26_033.cs
alyscs: Fixed via ThunderG99/DCGO#1: CanUseCondition changed from card.Owner.HandCards.Contains(card) to true — the hand-check was preventing the condition from re-evaluating correctly, gating now happens fully inside ChangeCost via CardSourceCondition/RootCondition/PermanentsCondition.

alyscs:
Recreated as a standalone PR: #1939 (per updated guidance to treat each card as its own fresh PR rather than a companion PR into an existing branch). ThunderG99/DCGO#1 and this PR are superseded by #1939.

PR #1699, opened 2026-07-21
Create BT26_073

### LINKED ISSUE #1656: BT26-073: Aegiochusmon: Dark
Image

ThunderG99:
Resolves #1656

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Purple/BT26_073.cs
+++ b/Assets/Scripts/CardEffect/BT26/Purple/BT26_073.cs
@@ -1,7 +1,5 @@
-using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 
 // Aegiochusmon: Dark
 namespace DCGO.CardEffects.BT26
@@ -11,17 +9,6 @@ namespace DCGO.CardEffects.BT26
         public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
         {
             List<ICardEffect> cardEffects = new List<ICardEffect>();
-
-            #region Inherited Effect
-
-            #region Security A. +1
-            if (timing == EffectTiming.None)
-            {
-                cardEffects.Add(CardEffectFactory.ChangeSelfSAttackStaticEffect(changeValue: 1, isInheritedEffect: true, card: card, condition: null));
-            }
-            #endregion
-
-            #endregion
 
             #region Alt Digivolution
             if (timing == EffectTiming.None)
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Purple/BT26_073.cs
```diff
@@ -0,0 +1,310 @@
...
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Alt Digivolution
+            if (timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.ContainsCardName("Aegiomon");
```
x89rt2, line 18: equalscardname, double check the exact one in the IDE but contains allows partial matches

hooperk: It is indeed EqualsCardName

Assets/Scripts/CardEffect/BT26/Purple/BT26_073.cs
```diff
@@ -0,0 +1,310 @@
...
+                bool CanDeleteCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card)
+                        && permanent.TopCard.HasLevel
+                        && permanent.Level <= 5;
+                }
+                #endregion
+
+                bool hasPaidCost = false;
+
```
hooperk, line 110: Just in case, if (!CanDeleteSelf() && !CanReturnFromTrash) yield break;
so if the effect cannot be done at all, don't give the user a single option dialogue

Assets/Scripts/CardEffect/BT26/Purple/BT26_073.cs
alyscs: Fixed via ThunderG99/DCGO#2 (opened against this branch since I don't have write access to it directly): the thread had confirmed EqualsCardName was correct, but the code was never actually updated to match — it was still ContainsCardName. Changed now.

Assets/Scripts/CardEffect/BT26/Purple/BT26_073.cs
alyscs: Fixed via ThunderG99/DCGO#2: added the yield break guard right after CanDeleteSelf/CanReturnFromTrash are defined, before the cost-selection dialogue is built.

alyscs:
Recreated as a standalone PR: #1940 (per updated guidance to treat each card as its own fresh PR rather than a companion PR into an existing branch). ThunderG99/DCGO#2 and this PR are superseded by #1940.

PR #1702, opened 2026-07-23
Show effects for any card in trash or on field even if using Root.Custom

hooperk:


sollybr:
Resolves: #1703

sollybr (commented):
This doesn't seem to address showing the pending effects of the opponent to the selecting player but I'll carry out empirical tests if x8 can't get to it first

hooperk:
> This doesn't seem to address showing the pending effects of the opponent to the selecting player but I'll carry out empirical tests if x8 can't get to it first

The reason opponent's effects don't show is because the behaviour of this selector for Root.Trash is to show the trash of the selecting player. In order to allow the current player to select cards from their opponent's trash, such effects are written with Root.Custom and customRootList: card.Owner.Enemy.TrashCards. The change done here stops check the root and  instead cares about where the card actually is, so that it still works when Root.Custom is used

sollybr (changes requested):
Empirical tests show pending effects neither in the selection box, nor in the trash-from-field view
image
image

Cards tested: BT18 Millennium (Jogress conditions)
Stack deleted: Phantomon (+ digivolution line)

x89rt2 (changes requested):
According to above it doesn't work?

x89rt2 (approved):

PR #1706, opened 2026-07-28
RandomDeckSelect

JoshuaEdgel:
Adds a Random Deck Selector button to the deck list on the deck selection screens. It does NOT auto start searching for a game, it only triggers the onClick of a random deck, allowing the user to decide if they really want to use the random deck it selected.


sollybr (changes requested):
To implement the button it's necessary to modify the scene through unity editor.

Image

JoshuaEdgel:
Committed what I hope is the only missing file.

sollybr:
Could you also make it so it doesn't select invalid decks?

JoshuaEdgel:
I have made it now so it doesn't select invalid decks, it also checks to see if all decks are invalid. It is not optimal in the slightest, it checks EVERY time the random deck button is pressed. I have a way in my brain to make it cleaner, just don't think I'll have the time today to implement it. For now it works!

sollybr (approved):
Had no problem with my tests. Works as intended. @x89rt2

x89rt2 (approved):

PR #1711, opened 2026-08-03
Revert "Revert "First attempt at solving issues with copy effects""

x89rt2:
Reverts DCGO2/DCGO#1520

PR #1713, opened 2026-08-05
Copy Card Factory Patch

sollybr:
Factory with AddSkillClass to avoid engine conflicts
Patch for 1.17.4+
Continuity to: #1711, #1705, #1509

### DIFF
--- a/Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
+++ b/Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
@@ -1,60 +1,153 @@
 using System;
+using System.Collections;
 using System.Collections.Generic;
 
 public partial class CardEffectFactory
 {
     /// <summary>
-    /// Effect to copy through the effects of digivolution cards as effect of the top card. Done directly in CEntity_Effect.CardEffects for efficiency
+    /// Creates an AddSkillClass static effect that dynamically copies non-inherited effects 
+    /// from digivolution cards under this Permanent.
     /// </summary>
-    /// <param name="cardEffects">Reference to the list of effects for this card to add effects to</param>
-    /// <param name="timing">timing being queried</param>
-    /// <param name="card">Card with the copy effect</param>
-    /// <param name="canUseCondition">Condition to use this effect (Current turn, This digimon with X trait, etc.)</param>
-    /// <param name="cardCondition">Condition of the cards to copy from (gammamon in name, MachineDramon or Chaosdramon, etc.)</param>
-    /// <param name="effectCondition">Conditions on the effects to copy (Only On Plays, only activate effects, etc.)</param>
-    public static void CopyDigivolutionCardEffects(ref List<ICardEffect> cardEffects,
-                                                   EffectTiming timing, 
-                                                   CardSource card,
-                                                   bool isInheritedEffect = false,
-                                                   bool isLinkedEffect = false,
-                                                   Func<bool> canUseCondition = null,
-                                                   Func<CardSource, bool> cardCondition = null,
-                                                   Func<ICardEffect, bool> effectCondition = null
-                                                   )
+    public static void CopyDigivolutionCardEffects(
+        ref List<ICardEffect> cardEffects,
+        EffectTiming timing,
+        CardSource card,
+        bool isInheritedEffect = false,
+        bool isLinkedEffect = false,
+        Func<List<CardSource>, List<CardSource>> targetSources = null,
+        Func<Hashtable, bool> canUseCondition = null,
+        Func<Permanent, bool> permanentCondition = null,
+        Func<CardSource, bool> cardSourceCondition = null,
+        Func<CardSource, bool> cardCondition = null,
+        Func<ICardEffect, bool> effectCondition = null
+        )
     {
-        if (card.PermanentOfThisCard() == null) return;
-        if (canUseCondition != null && !canUseCondition()) return;
-
-        Permanent thisPermanent = card.PermanentOfThisCard();
-
-        //If it is an inherited effect it should be in digivolution cards, if link effect should be linked, if neither it should not be in either (leaving only topcard)
-        if (isInheritedEffect != thisPermanent.DigivolutionCards.Contains(card)) return;
-        if (isLinkedEffect != thisPermanent.LinkedCards.Contains(card)) return;
-
-        List<CardSource> validSources = thisPermanent.DigivolutionCards.Filter(cardSource => cardSource != card && (cardCondition == null || cardCondition(cardSource)));
-
-        foreach (CardSource cardSource in validSources)
+        
+        if (timing is not EffectTiming.None)
         {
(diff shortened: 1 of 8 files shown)


sollybr:
Leaving as draft until Bagra copies are good to go

sollybr:
Key motivation files:
https://github.com/DCGO2/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/MultipleSkills.cs#L8
https://github.com/DCGO2/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/AutoProcessing.cs#L9
https://github.com/DCGO2/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/Permanent.cs#L7

---

Important considerations:
https://github.com/DCGO2/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/CEntity_Effect.cs#L6
https://github.com/DCGO2/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/CEntity_EffectController.cs#L5
https://github.com/DCGO2/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/CardEffectCommons/GameContextDeterminarion.cs#L5

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
```diff
@@ -1,60 +1,153 @@
...
+                            hashtable => validSources(targetSources(null)).Contains(activateClass.OriginalEffectSourceCard) 
+                            && (originalUseCondition is null || originalUseCondition(hashtable))
+                        );
+
+                        var originalActivateCondition = activateClass.CanActivateCondition;
+                        activateClass.SetCanActivateCondition(
+                            hashtable => validSources(targetSources(null)).Contains(activateClass.OriginalEffectSourceCard) 
+                            && (originalActivateCondition is null || originalActivateCondition(hashtable))
+                        );
+
```
hooperk, line 120: We appear to be missing setting the hashstring, I believe we should still do that in here to separate different OPTs. What I did miss in my original is that the usage of i should have been i++ so it would be different per effect on a card.

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
```diff
@@ -1,60 +1,153 @@
...
+            getEffects: GetEffects,
+            limitTiming: null
+        );
+
+        cardEffects.Add(addSkillClass);
     }
 
-    private static string GenerateHashString(CardSource card, CardSource cardSource, int i) => $"Card-{card.CardIndex}-Copying-Card-{cardSource.CardIndex}-effect-{i}";
+    private static string GenerateHashString(CardSource card, CardSource cardSource, int i)
+        => $"Card-{card.CardIndex}-Copying-Card-{cardSource.CardIndex}-effect-{i}";
```
hooperk, line 152: It occurred to me just now, We should include -inherited and -linked when the effect comes from such an effect, to differentiate Canoweissmon own effect vs inherited effect

x89rt2 (changes requested):
check above

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
sollybr: you mean this?
https://github.com/sollybr/DCGO/blob/4d57360a8ef3bf3499630171cc38b3eb7752505f/Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs#L66-L69

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
sollybr: to avoid unexpected fizzles, i think it's best to let CardSource be the definitive discernment of effects:
https://github.com/sollybr/DCGO/blob/1055513dc3fb4d12a8f4d813662711160ea3264c/Assets/Scripts/Script/ICardEffect.cs#L892

sollybr:
Waiting for feedback on reviews

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
hooperk: No, this function which gives each copied effect a unique Hashstring based on card it is copied from to ensure correct OPT behaviour

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
hooperk: That is not how the sim checks Once per Turn usages, they work off the HashStrings to check if it is the same effect
https://github.com/DCGO2/DCGO/blob/5dfbb9436f61434ae77a72b58b73289048c00d45/Assets/Scripts/Script/ICardEffect.cs#L848

A copied effect needs to have it's own OPT for the copying effect, separate to the origin card's. 
Giving it a hashstring that identifies the source card uniquely would then be tracked in the cEntity_EffectContrioller of the card with the copy effect but still needs to be able to differentiate the OPT effects of 2 identical cards. That's what the GenerateHashString method below was aiming to do

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
sollybr: let's suppose one of OPT copied effects triggers for DkBagra
it trashes N cards under the permanent
activates and nothing else changes for the permanent
then there's another check for copied effects triggers; how is the arbitrary index `i` in the function guaranteed to give the effect that already activated the same hash?
```css
private static string GenerateHashString(CardSource card, CardSource cardSource, int i)
        => $"Card-{card.CardIndex}-Copying-Card-{cardSource.CardIndex}-effect-{i}";
```

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
```diff
@@ -1,60 +1,153 @@
...
+            return getCardEffects;
         }
+
+        addSkillClass.SetUpAddSkillClass(
+            cardSourceCondition: cardSourceCondition,
+            getEffects: GetEffects,
+            limitTiming: null
+        );
+
+        cardEffects.Add(addSkillClass);
```
hooperk, line 148: Since this is now just adding the 1 single AddSkillEffect, it could now just have return type of IAddSkillEffect, and the source cards could have 
```
if (timing == EffectTiming.None)
}
    cardEffects.Add(CardEffectFactory.CopyDigivolutionEffect(...));
}
```
Doesn't have to be changed now, but would make sense to change for Succession so the SuccessionSelfEffect matches similar ones rather than passing a ref to cardEffects through.

PR #1719, opened 2026-08-09
Don't reference permanent if deleted

hooperk:


Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
```diff
@@ -132,7 +150,7 @@ List<ICardEffect> GetEffects(CardSource sourceCard, List<ICardEffect> getCardEff
                         getCardEffects.Add(PermanentEffectFactory.AddDetailClass(
                             thisPermanent,
                             cardEffect.EffectDescription,
-                            true,
+                            false,
```
hooperk, line 153: Unrelated but saw while I was in here anyway. This boolean is "IsTriggeredEffect", which is essentially is activate class. So it is false for all other kinds of cardeffect. It isn't used yet but the potential is to display a warnign when a card has a granted effect that will trigger upon an action, like "[When Attacking] Lose 2 memory"

PR #1722, opened 2026-08-09
Update EX9_074.cs

x89rt2:
Modernized card, shared OP/WD, changed deletion to not bypass immunities, changed color count method to new Susan method. We still have an issue that the 6+ deletion method is logically flawed in how you select targets.

Update: OK second commit has a method that might work for the 6+ deletion. Needs checking.

### DIFF
--- a/Assets/Scripts/CardEffect/EX9/White/EX9_074.cs
+++ b/Assets/Scripts/CardEffect/EX9/White/EX9_074.cs
@@ -299,7 +299,11 @@ namespace DCGO.CardEffects.EX9
                         }
                     }
                 }
-            }
+                    if (permanentToDelete.Count > 0)
+{
+                            yield return ContinuousController.instance.StartCoroutine(
+                                new DestroyPermanentsClass(isNotImmune, CardEffectCommons.CardEffectHashtable(activateClass)).Destroy());
+}
             #endregion
 
             #region All Turns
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/EX9/White/EX9_074.cs
```diff
@@ -94,445 +81,254 @@ bool CanTargetCondition_ByPreSelecetedList(List<CardSource> cardSources, CardSou
...
-                        {
-                            Hashtable _hashtable = new Hashtable();
-                            hashtable.Add("CardEffect", activateClass);
-                            yield return ContinuousController.instance.StartCoroutine(new DestroyPermanentsClass(permanentToDelete, _hashtable).Destroy());
+                            yield return ContinuousController.instance.StartCoroutine(
+                                new DestroyPermanentsClass(isNotImmune, hashtable).Destroy());
                         }
                     }
                 }
             }
```
hooperk, line 290:
```suggestion
                    if (permanentToDelete.Count > 0)
{
                            yield return ContinuousController.instance.StartCoroutine(
                                new DestroyPermanentsClass(isNotImmune, CardEffectCommons.CardEffectHashtable(activateClass)).Destroy());
}
```

hooperk (changes requested):

Assets/Scripts/CardEffect/EX9/White/EX9_074.cs
hooperk: This attached to a weird place on the diff, but is what is needed instead of the whole change within permanentToDelete > 0 block

PR #1720, opened 2026-08-09
Copy Effects tidy up

hooperk:
Lower priority PR for some tidying up of the copy effect after rapid fire changes to get it working

PR #1723, opened 2026-08-10
Implement BT26-103 Jupitermon: Wrath Mode

### LINKED ISSUE #1686: BT26-103: Jupitermon: Wrath Mode
Image

hooperk:
Built over https://github.com/DCGO2/DCGO/pull/1720 , We may want to merge that first but merging this would just close both

Succession does most of the special logic in the Copied effect factory method, in order to be most efficient it handles stopping after the topmost card there. The CEF Succession method is then mostly just a mapping from our usual Static Keyword format to the Copy factory.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Yellow/BT26_103.cs
+++ b/Assets/Scripts/CardEffect/BT26/Yellow/BT26_103.cs
@@ -1,0 +1,157 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+// Jupitermon: Wrath Mode
+namespace DCGO.CardEffects.BT26
+{
+    public class BT26_103 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Alternate Digivolution Requirement
+            if (timing == EffectTiming.None)
+            {
+                static bool PermanentCondition(Permanent targetPermanent)
+                {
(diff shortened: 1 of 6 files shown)


Assets/Scripts/CardEffect/BT26/Yellow/BT26_103.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+
+            CardEffectFactory.ActivateClassesForSharedEffects
+                (ref cardEffects, timing, card,
+                    SharedEffectName,
+                    SharedActivateCoroutine,
+                    SharedEffectDescription,
+                    maxCountPerTurn: 1,
+                    hashValue: SharedHashString,
+                    optional: false,
```
hooperk, line 84: Something to call out here: Counter should be optional at the game level, and so I have not done anything to tinker with that for just this card. I want to test how the sim behaves for a non-optional Counter effect and this is the first to exist. If it does not behave I want to try fixing that in the core code first before just making us always make counter effects optional.

Things to test:
counter timing when this is the only effect and ensure you can choose not to activate it
counter timing while also have a valid blast in hand and ensure you can choose to do none of them

hooperk: If I cannot get it working in a core way, my plan then is to make it so the ActivateClassesForSharedEffects always sets counter effects as optional regardless of input for optional, and then any counter effects declared outside that method will just need to be optional. Putting this here in case someone else needs to implement that.

sollybr (approved):
Seems easy to read and fix. Can't do tests right now but should be good for first rounds of testing.

x89rt2 (approved):

PR #1726, opened 2026-08-11
End game view board

JoshuaEdgel:
Adds a button to the result screen allowing the user to hide the result screen and admire the board. Also adds a button to the board to bring the result screen back up. Nothing on the board is interactable (besides settings and returning to the result screen).


JoshuaEdgel:
I did notice that it leaves the options button and all the options inside of it intractable (including the surrender button), in my testing that doesn't seem to affect anything. But I wanted it known in case someone has a preference for how they want it to be.

x89rt2 (approved):

PR #1727, opened 2026-08-12
Turn Counter

JoshuaEdgel:
Adds a Turn counter to the Play Log to make it easier to tell when things happened.
image


hooperk (approved):

JoshuaEdgel:
I could have it say "Game Start" instead of "Turn: 0" if we think that makes more sense. I just stuck with the "Turn" to keep it uniform.

x89rt2 (approved):

PR #1730, opened 2026-08-14
Create BT26_005 Pinamon

alyscs:
## Summary
- Implements BT26-005 Pinamon's Inherited Effect: by trashing the bottom face-down card from under a Tamer, may play 1 play-cost-5-or-lower [Avian] or [DATA SQUAD] trait card from trash without paying the cost.
- Gates the free-play benefit on the trash actually succeeding, uses `ContainsTraits` instead of `EqualsTraits` for the trait check, and reuses `Permanent.HasFaceDownDigivolutionCards`.


x89rt2 (changes requested):
This is missing stuff like a canactivatecondition. Look at it again and let me know when it's ready for rereview.

alyscs:
Added the missing CanActivateCondition (was passing null) — uses CardEffectCommons.CanActivateOnDeletion, matching CanUseCondition's paired CanTriggerOnDeletion. Ready for re-review.

x89rt2 (approved):

PR #1732, opened 2026-08-14
Create BT26_001 Yokomon (+ new OnAddLibraryAnyone timing)

alyscs:
## Summary
- Implements BT26-001 Yokomon's Inherited Effect: when your effects add a card to the deck (top or bottom, from any origin - hand, security, field, or a digivolution source of any card), this Digimon may digivolve into a Digimon card with [Chronomon] in its text in hand with the cost reduced by 1.
- **Adds `EffectTiming.OnAddLibraryAnyone`**, a new engine-level timing, because no existing one covered this. `CardObjectController.AddLibraryTopCards`/`AddLibraryBottomCards` (the shared functions every card-to-deck movement already passes through) previously only broadcast `OnReturnCardsToLibraryFromTrash`, and only when the source was specifically the trash. This new timing is stacked unconditionally at the end of both functions, after the move has already happened (not a "would" pre-check - `WhenReturntoLibraryAnyone` is specifically a pre-action interceptor for whole-permanent deck bounces and doesn't fit this card).
- Also wired into `AddSecurityCard`'s Digi-Egg-to-DigitamaLibrary path, the one other place cards silently entered a library without broadcasting anything.
- Adds `CardEffectCommons.CanTriggerOnAddLibrary` (reuses the existing `GetCardSourcesFromHashtable` helper, matches the sibling `CanTriggerWhenDiscardLibrary` convention of excluding mid-reveal/null cards) and `IsAddLibraryTop` to read the new timing's hashtable. Only cards actually inserted (not tokens, not dedupe-skipped) are broadcast.

**Known limitation:** "your effects add..." is approximated as "cards added to *your* deck" (`cardSource.Owner == card.Owner`), since `AddLibraryTopCards`/`AddLibraryBottomCards` don't carry which player's effect caused the move - the existing `OnReturnCardsToLibraryFromTrash` timing has this same limitation. Properly tracking the causing effect would require plumbing an `ICardEffect` parameter through both functions and their ~80 call sites; flagging as a follow-up rather than in scope here.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
+++ b/Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
@@ -23,16 +23,13 @@ namespace DCGO.CardEffects.BT26
                 string EffectDescription()
                     => "[Your Turn] [Once Per Turn] When your effects add to decks, this Digimon may digivolve into a Digimon card with [Chronomon] in its text in the hand with the cost reduced by 1.";
 
-                // Approximates "your effects" by checking who owns the returned card, since
-                // AddLibraryTopCards/BottomCards don't carry which player's effect caused the
-                // move (same limitation the existing OnReturnCardsToLibraryFromTrash timing has).
-                bool CardSourceCondition(CardSource cardSource)
-                    => cardSource.Owner == card.Owner;
+                bool CardEffectCondition(ICardEffect cardEffect)
+                    => CardEffectCommons.IsOwnerEffect(cardEffect, card);
 
                 bool CanUseCondition(Hashtable hashtable)
                     => CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass)
                         && CardEffectCommons.IsOwnerTurn(card)
-                        && CardEffectCommons.CanTriggerOnAddLibrary(hashtable, CardSourceCondition);
+                        && CardEffectCommons.CanTriggerOnAddLibrary(hashtable, null, CardEffectCondition);
 
                 bool CanActivateCondition(Hashtable hashtable)
                     => CardEffectCommons.IsExistOnBattleAreaActivate(card, activateClass);
@@ -46,7 +43,7 @@ namespace DCGO.CardEffects.BT26
                         targetPermanent: card.PermanentOfThisCard(),
                         cardCondition: CardCondition,
                         payCost: true,
-                        reduceCostTuple: (1, CardCondition),
+                        reduceCostTuple: (1, null),
                         fixedCostTuple: null,
                         ignoreDigivolutionRequirementFixedCost: -1,
                         isHand: true,
--- a/Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
+++ b/Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
@@ -1,0 +1,60 @@
+using System.Collections;
+using System.Collections.Generic;
+
+// Yokomon
+namespace DCGO.CardEffects.BT26
+{
+    public class BT26_001 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Inherit
+            if (timing == EffectTiming.OnAddLibraryAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Digivolve into [Chronomon] card in hand, cost -1", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDescription());
+                activateClass.SetIsInheritedEffect(true);
+                activateClass.SetHashString("BT26_001_Inherit");
+                cardEffects.Add(activateClass);
+
+                string EffectDescription()
+                    => "[Your Turn] [Once Per Turn] When your effects add to decks, this Digimon may digivolve into a Digimon card with [Chronomon] in its text in the hand with the cost reduced by 1.";
+
+                bool CardSourceCondition(CardSource cardSource)
+                    => cardSource.Owner == card.Owner;
+
+                bool CanUseCondition(Hashtable hashtable)
+                    => CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass)
+                        && CardEffectCommons.IsOwnerTurn(card)
+                        && CardEffectCommons.CanTriggerOnAddLibrary(hashtable, CardSourceCondition);
+
+                bool CanActivateCondition(Hashtable hashtable)
+                    => CardEffectCommons.IsExistOnBattleAreaActivate(card, activateClass);
+
+                bool CardCondition(CardSource cardSource)
+                    => cardSource.IsDigimon && cardSource.HasText("Chronomon");
+
+                IEnumerator ActivateCoroutine(Hashtable _hashtable)
+                {
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DigivolveIntoHandOrTrashCard(
+                        targetPermanent: card.PermanentOfThisCard(),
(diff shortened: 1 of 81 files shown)


Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
```diff
@@ -0,0 +1,63 @@
...
+                activateClass.SetUpICardEffect("Digivolve into [Chronomon] card in hand, cost -1", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDescription());
+                activateClass.SetIsInheritedEffect(true);
+                activateClass.SetHashString("BT26_001_Inherit");
+                cardEffects.Add(activateClass);
+
+                string EffectDescription()
+                    => "[Your Turn] [Once Per Turn] When your effects add to decks, this Digimon may digivolve into a Digimon card with [Chronomon] in its text in the hand with the cost reduced by 1.";
+
+                // Approximates "your effects" by checking who owns the returned card, since
```
hooperk: Unfortunately this is an insufficient approximation, it needs to trigger for opponent's cards being returned to deck by the owners effects as well

hooperk: Perhaps it might be possible to look at where AddLibraryTopCards/BottomCards is called from as I believe ICardEffect should be accessible in most of those locations?

Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
```diff
@@ -0,0 +1,63 @@
...
+                bool CardCondition(CardSource cardSource)
+                    => cardSource.IsDigimon && cardSource.HasText("Chronomon");
+
+                IEnumerator ActivateCoroutine(Hashtable _hashtable)
+                {
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DigivolveIntoHandOrTrashCard(
+                        targetPermanent: card.PermanentOfThisCard(),
+                        cardCondition: CardCondition,
+                        payCost: true,
+                        reduceCostTuple: (1, CardCondition),
```
hooperk: Minor: CardCondition does not need to be repeated here, and it is a bit more efficient not to recheck for cost reduction, given the cardCondition is already applied as a filter on line 47

hooperk (changes requested):

hooperk:
I am also trying to check in with other judges on if adding cards to digi-egg deck should or should not trigger these effects.

hooperk:
One last thing to consider: Returning cards that were revealed from the deck back to the deck should not trigger these effects, currently any "reveal X, add to hand and return rest to bottom of deck" will trigger this because it uses AddLibraryBottomCards (and sometimes Top). We will need a way to not trigger in this case, or potentially to check the source location and ignore if it was already Library

Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
alyscs: Fixed — threaded an optional ICardEffect through AddLibraryTopCards/AddLibraryBottomCards into the OnAddLibraryAnyone hashtable (as "CardEffect", following the existing hashtable convention). BT26_001 now checks CardEffectCommons.IsOwnerEffect(cardEffect, card) instead of card ownership, so it triggers for opponent's cards returned by the owner's effects too.

Assets/Scripts/CardEffect/BT26/Red/BT26_001.cs
alyscs: Fixed — reduceCostTuple now passes null instead of re-checking CardCondition, since candidates are already filtered by it via the cardCondition param upstream.

alyscs:
Fixed the reveal-then-return false trigger too: AddLibraryTopCards/BottomCards now snapshot which cardSources were already in the library before the move (before RemoveFromAllArea runs) and exclude those from the addedCardSources passed to OnAddLibraryAnyone. So "reveal top X, add some to hand, return the rest to the deck" no longer fires this timing for the returned cards, since they never actually left the library.

On the digi-egg deck question — I left that open since you mentioned still checking with other judges; happy to adjust once that's settled.

Only Yokomon passes its ICardEffect through AddLibraryTopCards/BottomCards for now — the other ~99 existing call sites default to null and are unaffected. Threading cardEffect through all of them so every card that returns things to a deck can be attributed is a much bigger sweep; I held off on that until the ruling is settled, since the design might still change. Let me know if you'd like that swept now regardless.

Assets/Scripts/Script/CardObjectController.cs
```diff
@@ -829,6 +840,7 @@ public static IEnumerator AddLibraryTopCards(List<CardSource> cardSources, bool
                     if (!cardSource.Owner.DigitamaLibraryCards.Contains(cardSource))
                     {
                         cardSource.Owner.DigitamaLibraryCards.Insert(0, cardSource);
+                        if (!alreadyInLibraryCardSources.Contains(cardSource)) addedCardSources.Add(cardSource);
```
hooperk: The current ruling opinion is that it should not trigger from Digieggs. Comment out this line in both methods within the Digiegg check, that leaves an easy way for us to change it back if we get official confirmation.
Also, please do pass through ICardEffect in all call sites. While that will make this a large PR, it unfortunately needs to be to implement this correctly, and I'd rather take the time to review that entire change than do it piecemeal. Even include callsites that are only returning digi-egg, the extra variable will just do nothing if the ruling remains as is, but will future proof in case the ruling changes.

hooperk (changes requested):

alyscs:
Pushed both parts of this ruling:

- Commented out (not removed) the `addedCardSources.Add(cardSource)` line inside the DigiEgg branch of both `AddLibraryTopCards` and `AddLibraryBottomCards` in `CardObjectController.cs`, with a note explaining why — easy to restore if the ruling changes later.
- Threaded `ICardEffect cardEffect` through every call site of `AddLibraryTopCards`/`AddLibraryBottomCards` across the codebase (previously only this effect passed it), including sites that only ever add DigiEggs.

Assets/Scripts/Script/CardObjectController.cs
alyscs: Pushed both parts of this ruling:

Commented out (not removed) the addedCardSources.Add(cardSource) line inside the DigiEgg branch of both AddLibraryTopCards and AddLibraryBottomCards in CardObjectController.cs, with a note explaining why — easy to restore if the ruling changes later.
Threaded ICardEffect cardEffect through every call site of AddLibraryTopCards/AddLibraryBottomCards across the codebase (previously only this effect passed it), including sites that only ever add DigiEggs.

Assets/Scripts/Script/CardObjectController.cs
```diff
@@ -856,14 +870,24 @@ public static IEnumerator AddLibraryTopCards(List<CardSource> cardSources, bool
...
         }
         #endregion
+
+        yield return ContinuousController.instance.StartCoroutine(FireOnAddLibraryAnyone(addedCardSources, isTop: true, cardEffect: cardEffect));
     }
     #endregion
 
     #region put a card at the bottom of the deck
-    public static IEnumerator AddLibraryBottomCards(List<CardSource> cardSources, bool notAddLog = false)
+    public static IEnumerator AddLibraryBottomCards(List<CardSource> cardSources, bool notAddLog = false, ICardEffect cardEffect = null)
```
hooperk: One final comment, can we remove the default value of null for both methods so we cannot forget to pass through the cardEffect going forward

hooperk (approved):
Looks good, I've 1 last change I'd like done for our own sanity, mentioned below

alyscs:
Removed the default null for cardEffect on both AddLibraryTopCards/AddLibraryBottomCards, so it's now a required argument everywhere. Only one non-CardEffect call site needed updating — TurnStateMachine's mulligan redraw, now passing cardEffect: null explicitly since no card effect causes that reshuffle.

PR #1734, opened 2026-08-15
Create BT26_007 Swipemon

alyscs:
## Summary
- Implements BT26-007 Swipemon's Inherited Effect: may link 1 [Seven Code] trait Digimon card from your hand or this Digimon's digivolution cards to this Digimon with the cost reduced by 2.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_007.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_007.cs
@@ -1,0 +1,146 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+// Swipemon
+namespace DCGO.CardEffects.BT26
+{
+    public class BT26_007 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Inherit
+            if (timing == EffectTiming.OnAllyAttack)
+            {
+                ActivateClass activateClass = new ActivateClass();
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Green/BT26_007.cs
```diff
@@ -0,0 +1,148 @@
...
+                        selectionElements.Add(new SelectionElement<int>(message: "From this Digimon's digivolution cards", value: 2, spriteIndex: 0));
+                    selectionElements.Add(new SelectionElement<int>(message: "Do not link", value: 3, spriteIndex: 1));
+
+                    GManager.instance.userSelectionManager.SetIntSelection(
+                        selectionElements: selectionElements,
+                        selectPlayer: card.Owner,
+                        selectPlayerMessage: "From which area will you link a card?",
+                        notSelectPlayerMessage: "The opponent is choosing from which area to select a card.");
+
+                    yield return ContinuousController.instance.StartCoroutine(GManager.instance.userSelectionManager.WaitForEndSelect());
```
x89rt2: Finish the isUsed set up, I've already changed the optional to false and added the isskippable. Also, make it so if there is only legal targets on 1 of the 2 locations, it skips the selection here, goes to the proper selection below with the canselectno as true with the isUsed at the correct locations and the reset the activation if not used at the end.

x89rt2 (changes requested):

alyscs:
Finished the isUsed setup (set true on an actual link, RemoveUse() otherwise) and added the skip-the-menu behavior — when only 1 of hand/digivolution-cards has a legal target, it goes straight to that location's own selection (canSelectNo already true there) instead of showing the 3-way area menu.

x89rt2 (approved):

PR #1755, opened 2026-08-17
Add BT26-030 Pumpkinmon card effect

alyscs:
## Summary
- Implements BT26-030 Pumpkinmon's card effect.
- Alternate digivolution requirement: Lv.4 w/[TS] trait, cost 3.
- [Security] You may play 1 card with the [Angel] or [TS] trait and a play cost of 4 or less from your hand or trash without paying the cost.
- [On Play] [When Digivolving] By trashing 1 card in your hand, 1 of your Digimon with the [Iliad] trait gains <Execute> and <Ascension> for the turn.
- No `CardEffectCommons.GainAscension` helper exists yet (only the self-targeted `AscensionSelfEffect`), so the targeted grant is built inline in this file mirroring `GainExecute`'s shape — `CanTriggerAscension`/`CanActivateAscension`/`AscensionProcess` already accept an arbitrary `CardSource`, so no shared engine changes were needed. Flagging this as a novel composition for review.


x89rt2:
@hooperk Needs gain ascension code.

x89rt2 (approved):

PR #1757, opened 2026-08-17
Add BT26-032 Ceresmon // Famis card effect

alyscs:
## Summary
- Implements BT26-032 Ceresmon // Famis's DUAL Digimon/Option card effect.
- Digimon side: alternate digivolution requirement — a play cost 12 [Ceresmon] card, cost 2. Alliance, Succession <<[Ceresmon]>>.
- [When Digivolving] All of your opponent's suspended Digimon get -5000 DP until their turn ends. Then, by suspending 1 Digimon, if it's your turn, you may play or use 1 card with the [Vegetation] or [TS] trait from your hand with the cost reduced by 5.
- Option side (Famis): Use Req. <<[TS] trait>>, [Main] you may suspend 2 of your opponent's Digimon or Tamers, then 3 of their Digimon or Tamers can't unsuspend until their turn ends, Arts Digivolve.
- The "by suspending 1 Digimon" cost on the Digimon side isn't explicitly scoped to "your Digimon" in the printed text — implemented as the card owner's own Digimon, matching how self-inflicted costs are conventionally worded on other cards. Flagging in case the intended scope is broader.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_032.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_032.cs
@@ -62,8 +62,7 @@ namespace DCGO.CardEffects.BT26
 
                 bool CanSelectHandCardCondition(CardSource cardSource)
                     => (cardSource.EqualsTraits("Vegetation") || cardSource.HasTSTraits)
-                        && ((cardSource.IsOption && !cardSource.CanNotPlayThisOption)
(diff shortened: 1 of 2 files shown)


alyscs:
Switched to CardEffectCommons.CanPlayOrUse for the Vegetation/TS play-or-use check, replacing the hand-rolled IsOption/HasPlayCost condition.

x89rt2 (approved):

PR #1762, opened 2026-08-17
Add BT26-038 Kuwagamon card effect

### LINKED ISSUE #1621: BT26-038: Kuwagamon
Image

alyscs:
## Summary
- Implements BT26-038 Kuwagamon's card effect.
- Alternate digivolution requirement: Lv.3 w/[TS] trait, cost 2.
- [When Moving] [On Play] [When Digivolving] You may suspend 1 Digimon. Then, 1 of your Digimon with the [Insectoid] or [Titan] trait gets +3000 DP until your opponent's turn ends.
- Inherited [Your Turn] [Once Per Turn] When this Digimon wins a battle, 1 of your Digimon with the [Insectoid] or [Titan] trait may digivolve into an [Insectoid] or [Titan] trait Digimon card in the hand with the cost reduced by 1.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_038.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_038.cs
@@ -39,9 +39,6 @@ namespace DCGO.CardEffects.BT26
                 => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
                     && (permanent.TopCard.ContainsTraits("Insectoid") || permanent.TopCard.ContainsTraits("Titan"));
 
-            bool SharedCanActivateCondition(Hashtable hashtable, ICardEffect activateClass)
-                => CardEffectCommons.IsExistOnBattleAreaActivate(card, activateClass);
-
             IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
             {
                 if (CardEffectCommons.HasMatchConditionPermanent(CanSelectSuspendCondition))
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Green/BT26_038.cs
```diff
@@ -0,0 +1,193 @@
...
+                {
+                    bool isUsed = false;
+
+                    Permanent selectedPermanent = null;
+
+                    SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+
+                    selectPermanentEffect.SetUp(
+                        selectPlayer: card.Owner,
+                        canTargetCondition: CanSelectTargetPermanentCondition,
```
hooperk: Same comment as Morphomon, we should be checking if the target can evolve, consolidate CanSelectTargetPermanentCondition and CanEvolveTargetPermanentCondition into a single condition and use throughout

Assets/Scripts/CardEffect/BT26/Green/BT26_038.cs
```diff
@@ -0,0 +1,193 @@
...
+                        afterSelectPermanentCoroutine: null,
+                        mode: SelectPermanentEffect.Mode.Tap,
+                        cardEffect: activateClass);
+
+                    selectSuspendEffect.SetUpCustomMessage("Select 1 Digimon to suspend.", "The opponent is selecting 1 Digimon to suspend.");
+
+                    yield return ContinuousController.instance.StartCoroutine(selectSuspendEffect.Activate());
+                }
+
+                SelectPermanentEffect selectBuffEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
```
hooperk, line 64: Just to be safe in case of weirdness, or some future effect blanking traits, wrap this second effect in an if ( exists CanSelectBuffTargetCondition) block as well, even though the card itself currently would always meet the condition

hooperk (changes requested):

x89rt2 (approved):

PR #1765, opened 2026-08-17
Add BT26-042 Okuwamon card effect

alyscs:
## Summary
- Implements BT26-042 Okuwamon's card effect.
- Alternate digivolution requirement: Lv.4 w/[Insectoid]/[TS] trait, cost 3.
- [On Play] [When Digivolving] Suspend 1 of your opponent's Digimon or Tamers. Then, 1 of their Digimon or Tamers can't unsuspend until their turn ends.
- [On Play] [When Attacking] [Once Per Turn] Until your opponent's turn ends, 1 of your Digimon with the [Insectoid] or [Titan] trait gains <Piercing> and +3000 DP.
- Inherited [All Turns] [Once Per Turn] When this Digimon deletes your opponent's Digimon in battle, trash their top security card.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_042.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_042.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections;
 using System.Collections.Generic;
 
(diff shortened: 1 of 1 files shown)


alyscs:
Fixed the same additionalActivateCondition gap as BT26_040 (Drimogemon) on ability A — two independent branches (suspend a target, grant can't-unsuspend), but the gate only checked for a suspend target. Now checks either. Ability B was already single-branch and fine.

x89rt2 (approved):

PR #1768, opened 2026-08-17
Add BT26-040 Drimogemon card effect

alyscs:
## Summary
- Implements BT26-040 Drimogemon's card effect.
- Training, Piercing.
- [When Moving] [On Play] Suspend 1 of your opponent's Digimon. Then, by placing 1 card in your hand face down as this Digimon's bottom digivolution card, this Digimon gets +1000 DP until your opponent's turn ends for each of its face-down digivolution cards.
- Inherited Piercing.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_040.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_040.cs
@@ -1,0 +1,158 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+// Drimogemon
+namespace DCGO.CardEffects.BT26
+{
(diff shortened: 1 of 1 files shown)


x89rt2 (changes requested):
Missing alt evo. Does not need additional activate condition unless you add to it that "or have a card in hand"

alyscs:
Both fixed:

- Added the missing alt digivolve: Lv.3 w/[DM] trait, cost 2.
- SharedAdditionalActivateCondition now checks for a valid suspend target OR a hand card, instead of only the suspend target — the two are independent branches of the same shared ability, so requiring the suspend target blocked activation on the "place a card face down" branch when no opponent's Digimon was suspendable.

x89rt2 (approved):

PR #1769, opened 2026-08-18
Add BT26-045 GranKuwagamon card effect

### LINKED ISSUE #1628: BT26-045: GranKuwagamon
Image

alyscs:
## Summary
- Implements BT26-045 GranKuwagamon's card effect.
- Alternate digivolution requirement: Lv.5 w/[Insectoid]/[TS] trait, cost 3.
- When this card would be played, if your hand has fewer cards than your opponent's, reduce the cost by 4.
- [On Play] [When Digivolving] [When Attacking] [Once Per Turn] You may play 1 level 4 or lower Digimon card with the [Insectoid] or [Titan] trait from your hand without paying the cost.
- [Your Turn] All of your Digimon with the [Insectoid] or [Titan] trait gain <Alliance>, <Piercing> and <Vortex>.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
@@ -88,9 +88,8 @@ namespace DCGO.CardEffects.BT26
                     && (cardSource.ContainsTraits("Insectoid") || cardSource.ContainsTraits("Titan"))
                     && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
 
-            bool SharedCanActivateCondition(Hashtable hashtable, ICardEffect activateClass)
-                => CardEffectCommons.IsExistOnBattleAreaActivate(card, activateClass)
-                    && CardEffectCommons.HasMatchConditionOwnersHand(card, cardSource => CanSelectHandCardCondition(cardSource, activateClass));
+            bool SharedAdditionalActivateCondition(Hashtable hashtable, ActivateClass activateClass)
+                => CardEffectCommons.HasMatchConditionOwnersHand(card, cardSource => CanSelectHandCardCondition(cardSource, activateClass));
 
             IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
             {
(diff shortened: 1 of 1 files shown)


x89rt2:
@hooperk need a check on the your turn

Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
```diff
@@ -0,0 +1,177 @@
...
+                            cost -= 4;
+                        }
+
+                        return cost;
+                    }
+
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.ShowReducedCost(_hashtable));
+                }
+            }
+            #endregion
```
hooperk, line 32: We have a newer way of doing this when the effect is mandatory and it ensures the cost is shown in hand to the player (though not applied for considerign the cost of the card for playing for free), CardEffectFactory.MandatorySelfPlayCostReduction. See AD1_018.cs for example of its use.

Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                    if (_timing == EffectTiming.OnEndTurn)
+                    {
+                        cardEffects.Add(CardEffectFactory.VortexSelfEffect(isInheritedEffect: false, card: cardSource, condition: GrantCondition));
+                    }
+
+                    return cardEffects;
+                }
+            }
+            #endregion
```
hooperk, line 134: We've had performance issues with doing this in the past. Do only Vortex with a limit timing above, since it needs to trigger separately for each digimon at EOT, and use AllianceStaticEffect and PierceEffect with above PermanentConditions

Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
```diff
@@ -0,0 +1,177 @@
...
+                onPlay: true,
+                whenDigivolving: true,
+                whenAttacking: true);
+
+            #region Your Turn - Grant Alliance/Piercing/Vortex
+            if (timing == EffectTiming.None)
+            {
+                AddSkillClass addSkillClass = new AddSkillClass();
+                addSkillClass.SetUpICardEffect("Your [Insectoid]/[Titan] Digimon gain Alliance, Piercing and Vortex", CanUseCondition, card);
+                addSkillClass.SetUpAddSkillClass(cardSourceCondition: CardSourceCondition, getEffects: GetEffects);
```
hooperk: Further to my comment below, 
```suggestion
                addSkillClass.SetUpAddSkillClass(cardSourceCondition: CardSourceCondition, getEffects: GetEffects, limitTiming: OnEndTurn);
```

hooperk (changes requested):

Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
alyscs: Done — switched to CardEffectFactory.MandatorySelfPlayCostReduction(4, card, Condition), matching AD1_018.cs.

Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
alyscs: Done — Alliance now uses AllianceStaticEffect(permanentCondition, ...) directly, and Vortex kept on AddSkillClass with limitTiming: EffectTiming.OnEndTurn per your suggestion below. For Pierce: there's no PierceStaticEffect(permanentCondition, ...) helper (only the single-target PierceEffect/PierceSelfEffect exist, no existing card calls the bare PierceEffect for a non-self target), so I registered it as a loop over currently-matching permanents under EffectTiming.OnDetermineDoSecurityCheck, calling PierceEffect(targetPermanent: permanent, ...) once per match each time CardEffects() is evaluated at that timing. Flagging this specifically since I don't have a real precedent to confirm this is the intended shape — let me know if there's a different pattern you had in mind.

Assets/Scripts/CardEffect/BT26/Green/BT26_045.cs
alyscs: Done — limitTiming: EffectTiming.OnEndTurn added, AddSkillClass now only handles Vortex.

x89rt2 (approved):

PR #1793, opened 2026-08-19
Add BT26-074 Cerberusmon card effect

alyscs:
## Summary
- Implements BT26-074 Cerberusmon's card effect.
- Alternate digivolution requirement: Lv.4 w/[TS] trait, cost 3.
- [On Play] [When Digivolving] [When Attacking] [Once Per Turn] If it's your turn, by trashing 1 card in your hand, you may use 1 Option card with the [Titan] trait from your trash with the cost reduced by 2.
- Inherited [On Deletion] Delete 1 of your opponent's Digimon with the lowest level.


alyscs:
Switched to CardEffectCommons.CanPlayOrUse for the Titan Option check — the previous CanPlayAsNewPermanent call was missing isPlayOption: true, so this ability was always dead code for any Option candidate.

x89rt2 (approved):

PR #1799, opened 2026-08-19
Add BT26-082 Ravemon card effect

alyscs:
## Summary
- Implements BT26-082 Ravemon's card effect.
- Alternate digivolution requirement: Lv.5 w/[DATA SQUAD] trait, cost 3.
- [Security] [End of Opponent's Turn] Play this card without paying the cost.
- [When Digivolving] [End of Attack] By deleting this Digimon or trashing 2 bottom face-down cards from under any of your Tamers, delete 1 of your opponent's highest DP Digimon.
- [On Deletion] Your opponent trashes 1 card in their hand. Then, if their hand has 7 or fewer cards, you may place this card face up as the bottom security card.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Purple/BT26_082.cs
+++ b/Assets/Scripts/CardEffect/BT26/Purple/BT26_082.cs
@@ -17,7 +17,7 @@ namespace DCGO.CardEffects.BT26
             {
                 static bool PermanentCondition(Permanent targetPermanent)
                 {
-                    return targetPermanent.TopCard.ContainsTraits("DATA SQUAD");
+                    return targetPermanent.TopCard.EqualsTraits("DATA SQUAD");
                 }
 
                 cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 3, ignoreDigivolutionRequirement: false, card: card, condition: null, level: 5));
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Purple/BT26_082.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            string SharedEffectName() => "By deleting this Digimon or trashing 2 Tamer's bottom face-down cards, delete 1 opponent's highest DP Digimon";
+
+            string SharedEffectDescription(string tag)
+                => $"[{tag}] By deleting this Digimon or trashing 2 bottom face-down cards from under any of your Tamers, delete 1 of your opponent's highest DP Digimon.";
+
+            bool IsTamerWithEnoughFaceDownCards(Permanent permanent)
+                => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaTamer(permanent, card)
+                    && permanent.DigivolutionCards.Count(cs => cs.IsFaceDown) >= 2;
+
```
x89rt2: same issue as bear cat, you can trash from 2 tamers

x89rt2 (changes requested):

alyscs:
Fixed — same bug shape as NightChiropmon. "any of your Tamers" means the 2 face-down cards can be split across up to 2 different Tamers (1 each), not just found on a single Tamer with 2. Replaced the single-Tamer selection with the same multi-select pattern used on BT26_070 (canEndSelectCondition auto-ends at 2 picks or when the first pick alone already has 2+ face-down cards; afterSelectPermanentCoroutine trashes 2 from one or 1 from each of two).

x89rt2 (approved):

PR #1801, opened 2026-08-20
Add BT26-084 Copipemon card effect

alyscs:
## Summary
- Implements BT26-084 Copipemon's card effect.
- Alternate digivolution requirement: Lv.2 w/[Appmon] trait, cost 0.
- Detach<<[Seven Code] trait>> (when this Digimon would leave the battle area other than by your effects, by trashing 1 of its specified link cards, it doesn't leave).
- [Your Turn] [Once Per Turn] When this Digimon gets linked, reveal the top 3 cards of your deck. You may play or use 1 [Seven Code] trait card among them with the cost reduced by 3. Return the rest to the top or bottom of the deck.
- Link [Appmon] trait: Cost 3.
- [When Linking] You may link 1 non-white level 4 or lower [System] or [Seven Code] trait card from your trash to this Digimon without paying the cost.

### DIFF
--- a/Assets/Scripts/CardEffect/BT21/Purple/BT21_073.cs
+++ b/Assets/Scripts/CardEffect/BT21/Purple/BT21_073.cs
@@ -23,7 +23,7 @@ namespace DCGO.CardEffects.BT21
             }
             #endregion
 
-            #region App Fusion (Globemon & Charismon)
+            #region App Fusion (Sociamon & Gossipmon)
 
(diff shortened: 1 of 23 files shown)


x89rt2 (changes requested):
Detach needs to be a key word.

alyscs:
> Detach needs to be a key word.

fixed

x89rt2 (approved):

PR #1814, opened 2026-08-20
Add BT26-099 Training Manual card effect

### LINKED ISSUE #1682: BT26-099: Training Manual
Image

alyscs:
## Summary
- Implements BT26-099 Training Manual's card effect.
- Use Req. [DM] trait.
- [Main] Reveal the top 3 cards of your deck. Add 1 [DM] trait card among them to the hand. Return the rest to the bottom of the deck. Then, place this card in the battle area.
- [All Turns] When face-down cards are placed in any of your Digimon's digivolution cards, <Delay>. Any of those Digimon may digivolve into a level 6 or lower [DM] trait Digimon card in the hand without paying the cost.
- [Security] Activate this card's [Main] effects.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Green/BT26_099.cs
+++ b/Assets/Scripts/CardEffect/BT26/Green/BT26_099.cs
@@ -88,7 +88,31 @@ namespace DCGO.CardEffects.BT26
 
                 IEnumerator ActivateCoroutine(Hashtable _hashtable)
                 {
-                    Permanent triggeringPermanent = CardEffectCommons.GetPermanentFromHashtable(_hashtable);
+                    List<Permanent> triggeringPermanents = new List<Permanent>();
+
+                    void AddTriggeringPermanent(Permanent permanent)
+                    {
+                        if (permanent != null && !triggeringPermanents.Contains(permanent))
+                        {
+                            triggeringPermanents.Add(permanent);
+                        }
+                    }
+
+                    AddTriggeringPermanent(CardEffectCommons.GetPermanentFromHashtable(_hashtable));
+
+                    // Other Digimon that had face-down cards placed on them by the same effect, in the same
+                    // window, before this trigger got resolved (e.g. an effect that places 1 card face down under
+                    // each of 2 selected Digimon at once) — each Digimon fires its own OnAddDigivolutionCards
+                    // hashtable, so gather any still-pending sibling triggers for this same card's ability.
+                    foreach (SkillInfo skillInfo in GManager.instance.autoProcessing.StackedSkillInfos)
+                    {
+                        if (skillInfo.Timing == EffectTiming.OnAddDigivolutionCards
+                            && skillInfo.CardEffect != null
+                            && skillInfo.CardEffect.EffectSourceCard == card)
+                        {
+                            AddTriggeringPermanent(CardEffectCommons.GetPermanentFromHashtable(skillInfo.Hashtable));
+                        }
+                    }
 
                     yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DeletePeremanentAndProcessAccordingToResult(
                         targetPermanents: new List<Permanent>() { card.PermanentOfThisCard() },
@@ -98,10 +122,12 @@ namespace DCGO.CardEffects.BT26
 
                     IEnumerator SuccessProcess(List<Permanent> permanents)
                     {
-                        if (triggeringPermanent != null && CanSelectDigivolveTargetCondition(triggeringPermanent))
+                        List<Permanent> eligiblePermanents = triggeringPermanents.Filter(CanSelectDigivolveTargetCondition);
+
+                        if (eligiblePermanents.Count == 1)
                         {
                             yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DigivolveIntoHandOrTrashCard(
-                                triggeringPermanent,
+                                eligiblePermanents[0],
                                 CanSelectDMCardCondition,
                                 payCost: false,
                                 reduceCostTuple: null,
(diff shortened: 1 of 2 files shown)


x89rt2 (changes requested):
Delay does not work correctly as is. Needs to captured the trigger digimon so only those digimon can be targets for the digivolve. I feel like I recently worked on a card with a similar thing, but I don't remember which atm.

alyscs:
Good catch — you're right, the target pool wasn't restricted to the triggering Digimon at all, it was matching any of my Digimon with a face-down digivolution card regardless of when/why it got there.

Fixed: the \`OnAddDigivolutionCards\` hashtable already carries the specific \`Permanent\` that received the face-down card(s), so I capture that directly via \`CardEffectCommons.GetPermanentFromHashtable(_hashtable)\` and digivolve off it, instead of re-scanning the whole battle area. Also dropped the \`SelectPermanentEffect\` selection step since there's only ever one legal target per trigger instance now.

x89rt2 (changes requested):
Still needs selection. There could be an effect that places cards under 2 digimon at same time

alyscs:
Confirmed there's no existing card that does this yet, so the goal is: if an effect ever places face-down cards under 2+ of my Digimon in one action, the player should get to pick which one digivolves.

Since OnAddDigivolutionCards fires once per Digimon (one Permanent per hashtable), that scenario means 2+ separate pending triggers for this same Training Manual instance stacking up before resolution. Added a scan of GManager.instance.autoProcessing.StackedSkillInfos for sibling OnAddDigivolutionCards triggers belonging to this same card (not yet resolved), and union their Permanents with the one from this trigger's own hashtable. Digivolves directly when there's exactly one eligible target, offers a SelectPermanentEffect scoped to just that set when there's more than one.

I don't have a precedent for this in the codebase, so flagging in case there's an established pattern I should be using instead — happy to adjust.

alyscs:
Refactored per your BT21_083 pointer. Couldn't reuse GetHashtablesFromHashtable directly though — that key only exists on hashtables built by OnEnterFieldHashtable/OnDeletionHashtable, where the caller already has every affected Permanent in hand before pushing. OnAddDigivolutionCards fires per-Permanent from Permanent's own instance method, called independently — so there's no single point that can pre-batch a "hashtables" list before the fact; siblings from the same still-pending window can only be discovered at check time.

Added CardEffectCommons.GetSimultaneousPermanentsFromAddDigivolutionCardsHashtable (in GetFromHashtable.cs, next to GetPermanentFromHashtable) that scans GManager.instance.autoProcessing.StackedSkillInfos for other not-yet-resolved OnAddDigivolutionCards entries and unions their Permanents in. BT26_099 now calls that instead of the inline scan, so it's reusable if another card needs this timing later.

Assets/Scripts/CardEffect/BT26/Green/BT26_099.cs
```diff
@@ -0,0 +1,179 @@
...
+                                eligiblePermanents[0],
+                                CanSelectDMCardCondition,
+                                payCost: false,
+                                reduceCostTuple: null,
+                                fixedCostTuple: null,
+                                ignoreDigivolutionRequirementFixedCost: -1,
+                                isHand: true,
+                                activateClass: activateClass,
+                                successProcess: null
+                            ));
```
hooperk, line 121: This could be replaced with selectedPermanent = eligiblePermanents[0], and then this block could exist just once outside of the below else

Assets/Scripts/Script/CardEffectCommons/GetFromHashtable.cs
```diff
@@ -726,6 +726,44 @@ public static Permanent GetPermanentFromHashtable(Hashtable hashtable)
...
+    /// instance method (Permanent.AddDigivolutionCards/AddDigivolutionCardsBottom), one hashtable per affected
+    /// Digimon, each pushed independently rather than pre-batched by a single caller that already knows every
+    /// affected Permanent up front. So this hashtable's own "Permanent" can't be batched at construction time —
+    /// it has to be gathered here, at check/resolution time, once any sibling pushes from the same still-pending
+    /// window have already been queued. Returns this hashtable's own Permanent plus any other Permanents with a
+    /// still-pending OnAddDigivolutionCards trigger in GManager.instance.autoProcessing.StackedSkillInfos (e.g. an
+    /// effect that places 1 face-down card under each of 2 selected Digimon in the same action) — callers should
+    /// still apply their own eligibility filter (existence/ownership/face-down checks) to the result.
+    /// </summary>
+    public static List<Permanent> GetSimultaneousPermanentsFromAddDigivolutionCardsHashtable(Hashtable hashtable)
```
hooperk: I'm not happy with this one. It could potentially pull in permanents not valid to the timing of the trigger. We also already are going to have multiple triggers since each triggered separately. Instead what we should do is have each trigger reference which permanent it relates to. We added functionality for that: ICardEffect.SetEffectTargets, with an example use in BT21_083.cs.
This function highlights which permanent(s) is related to the trigger in the case of multiple valid triggers for one effect. The effect will need to be set as optional:true to properly use it

hooperk (changes requested):

Assets/Scripts/CardEffect/BT26/Green/BT26_099.cs
alyscs: Superseded — switched this whole block to optional:true + SetEffectTargets per your other comment, so there's no longer a 1-vs-2+-eligible-permanents branch at all (each simultaneous trigger only ever has 1 target, its own hashtable's Permanent).

Assets/Scripts/Script/CardEffectCommons/GetFromHashtable.cs
alyscs: Switched to optional:true + activateClass.SetEffectTargets(TargetPermanents), following BT21_083.cs — each OnAddDigivolutionCards trigger now resolves independently against just GetPermanentFromHashtable(hashtable) for that specific instance, both in CanActivateCondition and ActivateCoroutine. Deleted GetSimultaneousPermanentsFromAddDigivolutionCardsHashtable entirely since nothing uses it anymore.

x89rt2 (approved):

PR #1816, opened 2026-08-20
Add BT26-101 Cross Arts card effect

### LINKED ISSUE #1684: BT26-101: Cross Arts
Image

alyscs:
## Summary
- Implements BT26-101 Cross Arts' card effect.
- Use Req. [TS] trait.
- [Main] If you have a Tamer with [Dan Yuki] or [Kanan Yuki] in its name, all of your [TS] trait Digimon gain Blocker and +3000 DP until your opponent's turn ends. Then, activate 1 of the effects below: delete 1 of your opponent's Digimon with as much DP as 1 of your [TS] trait Digimon or less, or 1 of your [TS] trait Digimon unsuspends.
- [Security] You may play 1 play cost 4 or lower [TS] trait card from your hand or trash without paying the cost.


Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
```diff
@@ -0,0 +1,208 @@
...
+
+                            GManager.instance.userSelectionManager.SetIntSelection(selectionElements: selectionElements, selectPlayer: card.Owner, selectPlayerMessage: "Select 1 effect to activate.", notSelectPlayerMessage: "The opponent is selecting 1 effect to activate.");
+                            yield return ContinuousController.instance.StartCoroutine(GManager.instance.userSelectionManager.WaitForEndSelect());
+
+                            choice = GManager.instance.userSelectionManager.SelectedIntValue;
+                        }
+                        else
+                        {
+                            choice = canSelectDelete ? 1 : 2;
+                        }
```
x89rt2: This section always needs to have both options available in case you specifically don't want to activate 1 of the effects.

Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
```diff
@@ -0,0 +1,208 @@
...
+                    {
+                        SelectCardEffect.Root root;
+
+                        if (canSelectHand && canSelectTrash)
+                        {
+                            List<SelectionElement<int>> selectionElements1 = new List<SelectionElement<int>>()
+                            {
+                                new(message: "From hand", value: 1, spriteIndex: 0),
+                                new(message: "From trash", value: 2, spriteIndex: 1),
+                                new(message: "Don't play", value: 3, spriteIndex: 2),
```
x89rt2:
```suggestion
                                new(message: "From hand", value: 1, spriteIndex: 0),
                                new(message: "From trash", value: 2, spriteIndex: 0),
                                new(message: "Don't play", value: 3, spriteIndex: 1),
```

x89rt2: This keeps showing up on some of the cards I've already fixed and merged. See if you can fix em on the rest of the PRs you have up.

x89rt2 (changes requested):

alyscs:
Swept all my open PRs for the spriteIndex issue. Found one real instance: BT26_100 (Dark Field, #1815) was using spriteIndex 0/1/2 for its hand/trash/decline menus — index 2 never matched anything in SelectCommand's fixed Buttons list, so "Don't play" never rendered. Fixed there (pushed).

Checked every other SelectionElement/Yes-No/decline-style selection across the other PRs and they already use the correct 0/1 pattern, so I don't believe there's anything else to fix on that front.

Now looking at BT26_101 itself for the two comments on this PR.

alyscs:
Fixed — the delete/unsuspend selection now always shows a "Don't activate either effect" option and always goes through the selection UI (only listing whichever effects are actually legal), instead of force-activating the single available effect when only one qualified.

(Also saw you already pushed the spriteIndex fix directly to this branch — thanks, that one's covered.)

x89rt2:
> Fixed — the delete/unsuspend selection now always shows a "Don't activate either effect" option and always goes through the selection UI (only listing whichever effects are actually legal), instead of force-activating the single available effect when only one qualified.
> 
> (Also saw you already pushed the spriteIndex fix directly to this branch — thanks, that one's covered.)

Incorrect, 1 of the 2 effects ALWAYS needs to activate, we just need to be able to choose either of the effects at any time, in case one of the effects is detrimental, we can use the other even if it would do nothing.

alyscs:
You're right, sorry — fixed properly this time. Both options are now always listed unconditionally (no decline option), and the unsuspend branch got the same existence guard the delete branch already had, so picking an option with no current legal target just does nothing instead of erroring or forcing a selection with zero candidates.

Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
```diff
@@ -0,0 +1,218 @@
...
+                        new(message: "Delete 1 of your opponent's Digimon with as much DP as 1 of your [TS] trait Digimon or less.", value: 1, spriteIndex: 0),
+                        new(message: "1 of your [TS] trait Digimon unsuspends.", value: 2, spriteIndex: 0),
+                    };
+
+                    GManager.instance.userSelectionManager.SetIntSelection(selectionElements: selectionElements, selectPlayer: card.Owner, selectPlayerMessage: "Select 1 effect to activate.", notSelectPlayerMessage: "The opponent is selecting 1 effect to activate.");
+                    yield return ContinuousController.instance.StartCoroutine(GManager.instance.userSelectionManager.WaitForEndSelect());
+
+                    int choice = GManager.instance.userSelectionManager.SelectedIntValue;
+
+                    int maxOwnTSDP = card.Owner.GetBattleAreaDigimons().Filter(OwnTSDigimonCondition).Map(p => p.DP).Max();
```
x89rt2: No, we have to be able to choose our digimon to use for our comparison in case we want to choose a weaker digimon to NOT delete an opponent's digimon.

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
alyscs: Understood, that matches what's already in the branch (your commit 5b62662f7 replaced the automatic .Max() with the manual comparison-Digimon selection for exactly this reason). No further change needed here.

x89rt2: Still seems incorrect. It looks like you are selecting your own digimon, but still are selecting opponent's based on your highest DP when it should be comparing to your own selected digimon's DP

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
alyscs: Traced the actual runtime behavior: the old code was already technically correct here — C# local functions capture outer variables by reference (not by value at declaration time), and `SelectPermanentEffect` invokes `canTargetCondition` live rather than caching it at `SetUp` time, so `CanSelectDeleteTargetCondition` was already reading the post-selection value by the time it actually ran.

That said, I get why it read as wrong — the auto-computed `.Max()` on one line, silently overwritten 30 lines later inside a nested coroutine, is genuinely confusing on a quick pass. Restructured to remove the ambiguity: replaced the auto-max seed with `int? compareDP = null` (a placeholder that can never accidentally match anything before your selection sets it), and dropped the pre-check gate entirely — the compare-Digimon selection now always appears when you pick "delete," consistent with the "always offer the choice even if it might no-op" principle from your earlier comment on this same PR.

Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
alyscs-azakaw: Fixed — addressed in a general PR comment at the time (the delete/unsuspend selection now always shows both options unconditionally, no decline point), but replying inline here too since that's the correct way to close this thread out.

Assets/Scripts/CardEffect/BT26/White/BT26_101.cs
alyscs-azakaw: Fixed — swept all my open PRs for this spriteIndex issue (posted the results as a general PR comment at the time). Found one real instance (BT26_100 Dark Field, #1815), fixed there. Every other selection across my PRs already used the correct 0/1 pattern. Replying inline here too since that's the correct way to close this thread out.

x89rt2 (approved):

PR #1817, opened 2026-08-20
Add BT26-102 Seven Code PAD card effect

alyscs:
## Summary
- Implements BT26-102 Seven Code PAD's card effect.
- Use Req. [Seven Code] trait.
- [Main] By placing 6 [Seven Code] trait Digimon cards from your battle area, link cards or trash as 1 of your [Seven Code] trait Digimon's bottom digivolution cards, that Digimon may digivolve into [Dantemon] in the hand, ignoring digivolution requirements and without paying the cost.
- [Security] You may play 1 play cost 5 or lower [Appmon] trait card from your hand or trash without paying the cost. Then, add this card to the hand.


x89rt2:
Main effect has issues. Can't have HasMatchConditionPermanent in can use. Needs a check on if has traited digimon in battle area at start of activatecoroutine, also needs a check that there even are 6 different traited digimon in total within all the correct locations. Also needs some kind of while loop that asks from which location to select cards in case you want to add cards in a certain order.

x89rt2 (changes requested):
Above comment

x89rt2:
Also security has same playbyeffect twice? Probably a way to only have it show once to make it cleaner.

alyscs:
Reworked Main: removed HasMatchConditionPermanent from CanUseCondition (the check now runs at the start of ActivateCoroutine), added an up-front check for whether 6 [Seven Code] cards are actually available across the battle area/link cards/trash before starting the gathering loop, and replaced the two sequential loops with a single loop where you pick which location to pull from each time (so order is your choice).

alyscs:
Also collapsed the two PlayByEffect call sites in Security into one — a shouldPlay flag + root variable are resolved first (via the 3-way choice when both hand and trash are available, or directly when only one is), then there's a single PlayByEffect call.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,280 @@
...
+
+                        if (locationElements.Count > 1)
+                        {
+                            GManager.instance.userSelectionManager.SetIntSelection(selectionElements: locationElements, selectPlayer: card.Owner, selectPlayerMessage: $"Select a location to take [Seven Code] digivolution material from ({remaining} remaining).", notSelectPlayerMessage: "The opponent is selecting a location for digivolution material.");
+                            yield return ContinuousController.instance.StartCoroutine(GManager.instance.userSelectionManager.WaitForEndSelect());
+                            selectedLocation = GManager.instance.userSelectionManager.SelectedIntValue;
+                        }
+                        else
+                        {
+                            selectedLocation = locationElements[0].value;
```
x89rt2: Error on this line. You have your IDE set up correctly?

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Confirmed and fixed — reproduced the CS1061 first (locationElements[0].value doesn't exist on SelectionElement<int>), then fixed it by computing selectedLocation directly from the canPickBattleArea/canPickLinked/canPickTrash flags instead, since that's how the single-item list was built anyway. Also set up a local Unity 2021.3.45f2 build so I can catch this class of error before pushing going forward — this branch was also missing 2 .meta files (BT26_102.cs.meta and the White folder's), fixed those too.

x89rt2 (changes requested):
This card has a bunch of issues with the Main effect and it's slowing me down trying to fix them all. I've done some of the work but I'm moving on for now.

alyscs:
Addressed both points from the Discord discussion, on top of the already-pushed fix for the link-card scoping:

1. The actual selection pool (customRootCardList) for the "link cards" option was still scoped to selectedTarget.LinkedCards only — inconsistent with the availability/canPick checks, which correctly count across all of your battle-area Digimon. Fixed the pool to match.
2. Nothing excluded already-selected link/trash cards from being picked again in a later loop iteration. Moved the relevant conditions inside ActivateCoroutine so they can check against materialCards and exclude anything already added. Battle area sources were already self-excluding since a selected permanent gets destroyed immediately.

Also found and fixed the "Cancel" option added in the link-card-scoping fix — the early-exit check and the selection handling for it were both backwards/missing, so Cancel wasn't actually functional. Fixed both.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,282 @@
...
+                            && permanent.LinkedCards.Filter(CanSelectLinkedOrTrashCondition).Count > 0;
+
+                    int availableBattleArea = CardEffectCommons.MatchConditionPermanentCount(CanSelectOtherBattleAreaCondition);
+                    int availableLinked = 0;
+                    int availableTrash = card.Owner.TrashCards.Filter(CanSelectLinkedOrTrashCondition).Count;
+
+                    foreach (Permanent permanent in card.Owner.GetBattleAreaDigimons().Where((permanent) => permanent.LinkedCards.Any(CanSelectLinkedOrTrashCondition)))
+                    {
+                        availableLinked += permanent.LinkedCards.Count(CanSelectLinkedOrTrashCondition);
+                    }
```
x89rt2, line 96: I started this but never finished it and can't do so now either. The issue with this one is it only counts each permanent as 1, when a guy with 2 linked cards would count as 2.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,282 @@
...
+                        bool canPickLinked = CardEffectCommons.HasMatchConditionPermanent(CanSelectLinkPermanentCondition);
+                        bool canPickTrash = card.Owner.TrashCards.Exists(CanSelectLinkedOrTrashCondition);
+
+                        List<SelectionElement<int>> locationElements = new List<SelectionElement<int>>();
+                        if (canPickBattleArea) locationElements.Add(new(message: "Battle area", value: 1, spriteIndex: 0));
+                        if (canPickLinked) locationElements.Add(new(message: "Link cards", value: 2, spriteIndex: 0));
+                        if (canPickTrash) locationElements.Add(new(message: "Trash", value: 3, spriteIndex: 0));
+                        locationElements.Add(new(message: "Cancel", value: 4, spriteIndex: 1));
+
+                        if (locationElements.Count == 1) yield break;
```
x89rt2: Unnecessary. The only way this could happen is if there were less than 6 targets before we started and that is taken care of above. No interruptive effect exists that would get between that and this.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,282 @@
...
+                        if (selectedLocation == 1)
+                        {
+                            Permanent selectedSource = null;
+
+                            SelectPermanentEffect selectSourceEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+
+                            selectSourceEffect.SetUp(
+                                selectPlayer: card.Owner,
+                                canTargetCondition: CanSelectOtherBattleAreaCondition,
+                                canTargetCondition_ByPreSelecetedList: null,
```
x89rt2: We need a canTargetCondition_ByPreSelecetedList so previous selected cards can't be reselected. This also needs to be referenced in lines 103-105.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,282 @@
...
+                                ? card.Owner.GetBattleAreaDigimons().SelectMany(permanent => permanent.LinkedCards).Where(CanSelectLinkedOrTrashCondition).ToList()
+                                : card.Owner.TrashCards.Filter(CanSelectLinkedOrTrashCondition);
+
+                            CardSource selectedCard = null;
+
+                            SelectCardEffect selectCardEffect = GManager.instance.GetComponent<SelectCardEffect>();
+
+                            selectCardEffect.SetUp(
+                                canTargetCondition: CanSelectLinkedOrTrashCondition,
+                                canTargetCondition_ByPreSelecetedList: null,
```
x89rt2: Needs canTargetCondition_ByPreSelecetedList

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,282 @@
...
+                            materialCards.Add(selectedSource.TopCard);
+                            remaining--;
+
+                            yield return ContinuousController.instance.StartCoroutine(new DestroyPermanentsClass(new List<Permanent>() { selectedSource }, CardEffectCommons.CardEffectHashtable(activateClass), notShowCards: true).Destroy());
+                        }
+                        else
+                        {
+                            List<CardSource> pool = selectedLocation == 2
+                                ? card.Owner.GetBattleAreaDigimons().SelectMany(permanent => permanent.LinkedCards).Where(CanSelectLinkedOrTrashCondition).ToList()
+                                : card.Owner.TrashCards.Filter(CanSelectLinkedOrTrashCondition);
```
x89rt2: I'm not sure this works? I think these need to be separate cause line 179 need to be root.trash or root.linkedcards

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,282 @@
...
+                    bool shouldPlay = canPlayFromHand || canPlayFromTrash;
+                    SelectCardEffect.Root root = canPlayFromTrash && !canPlayFromHand ? SelectCardEffect.Root.Trash : SelectCardEffect.Root.Hand;
+
+                    if (canPlayFromHand && canPlayFromTrash)
+                    {
+                        List<SelectionElement<int>> selectionElements = new List<SelectionElement<int>>()
+                        {
+                            new(message: "Play from hand", value: 1, spriteIndex: 0),
+                            new(message: "Play from trash", value: 2, spriteIndex: 1),
+                            new(message: "Don't play a card", value: 3, spriteIndex: 2),
```
x89rt2:
```suggestion
                            new(message: "Play from hand", value: 1, spriteIndex: 0),
                            new(message: "Play from trash", value: 2, spriteIndex: 0),
                            new(message: "Don't play a card", value: 3, spriteIndex: 1),
```

x89rt2: sprite index 0 is generally positive and 1 is generally negative

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Double-checked this one — it's actually correct as written. permanent.LinkedCards.Count(predicate) resolves to Enumerable.Count(predicate) here (this file has `using System.Linq;`), which does properly count every matching card per permanent, not just 1. The Count-property-shadowing trap (List<T>.Count is a property, so .Count(predicate) fails to compile as CS1955) only bites when System.Linq isn't in scope — ran into that exact thing on a different card (Zanbamon) earlier where the using was missing. Verified via a real Unity build that this compiles and traced the logic; left it as-is.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Removed — you're right, it's unreachable given the up-front availability check and no interruptive effect exists here. The Cancel option already covers the deliberate-exit case.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Added canTargetCondition_ByPreSelecetedList to this call (checking against materialCards) and to the two SelectCardEffect calls below, alongside the existing exclusion baked into CanSelectLinkedOrTrashCondition.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Added.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: You were right to be suspicious — split link cards and trash into two separate branches, each using its actual named Root (LinkedCards / Trash) instead of Root.Custom. RootCardList() always prefers customRootCardList when set, so the pool itself was fine either way, but CanSelectCard()'s flipped-card exclusion is specifically skipped for Root.Custom (along with Library/Security/DigivolutionCards) — meaning face-down linked/trash cards (e.g. cards placed face-down under a Tamer, a real mechanic elsewhere in this set) could have been selected as material under Root.Custom, which shouldn't be eligible. Fixed.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Applied.

alyscs:
Addressed the latest round of comments:
1. Verified the linked-card counting (line 95) is actually correct as-is — Enumerable.Count(predicate) resolves properly since System.Linq is imported in this file, so it does count multiple cards per permanent, not just 1.
2. Removed the now-unnecessary auto-break check per your note about the invariant.
3. Added canTargetCondition_ByPreSelecetedList to all three selection calls.
4. Split link-cards/trash into separate branches using their real named Root (LinkedCards/Trash) instead of Root.Custom — found a real gap doing this: Root.Custom skips the engine's flipped-card exclusion, so face-down link/trash cards could have been selected as material, which shouldn't be possible.
5. Applied the sprite-index suggestion in Security.

All verified against a real local Unity build before pushing.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+                        if (selectedLocation == 1)
+                        {
+                            Permanent selectedSource = null;
+
+                            SelectPermanentEffect selectSourceEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+
+                            selectSourceEffect.SetUp(
+                                selectPlayer: card.Owner,
+                                canTargetCondition: CanSelectOtherBattleAreaCondition,
+                                canTargetCondition_ByPreSelecetedList: (_, permanent) => !materialCards.Contains(permanent.TopCard),
```
hooperk: Consider having the materialCards check inside the PermanentCondition to match the behaviour of the other selections

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+                                canEndSelectCondition: null,
+                                maxCount: 1,
```
hooperk: maxCount of Math.Min(6-materialcards.Count, MatchConditionPermanentCount), allow multi select.
canEndNotMax: true

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+                                canNoSelect: () => false,
+                                selectCardCoroutine: SelectCardCoroutine,
+                                afterSelectCardCoroutine: null,
+                                message: $"Select 1 [Seven Code] trait Digimon card from your link cards to place as material ({remaining} remaining).",
+                                maxCount: 1,
+                                canEndNotMax: false,
+                                isShowOpponent: true,
+                                mode: SelectCardEffect.Mode.Custom,
+                                root: SelectCardEffect.Root.LinkedCards,
+                                customRootCardList: pool,
```
hooperk: I know it'd be nice to be able to pick things faster, but this hides from the player important information of which Digimon it is you are taking the Linked card from, which is why we don't do it this way. Select a single Permanent and then select from the Link Cards of that permanent

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+
+                            selectCardEffect.SetUp(
+                                canTargetCondition: CanSelectLinkedOrTrashCondition,
+                                canTargetCondition_ByPreSelecetedList: (_, cardSource) => !materialCards.Contains(cardSource),
+                                canEndSelectCondition: null,
+                                canNoSelect: () => false,
+                                selectCardCoroutine: SelectCardCoroutine,
+                                afterSelectCardCoroutine: null,
+                                message: $"Select 1 [Seven Code] trait Digimon card from your trash to place as material ({remaining} remaining).",
+                                maxCount: 1,
```
hooperk: maxCount of Math.Min(6-materialcards.Count, MatchConditionCardsInTrash), allow multi select.
canEndNotMax: true

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+
+                            selectCardEffect.SetUp(
+                                canTargetCondition: CanSelectLinkedOrTrashCondition,
+                                canTargetCondition_ByPreSelecetedList: (_, cardSource) => !materialCards.Contains(cardSource),
+                                canEndSelectCondition: null,
+                                canNoSelect: () => false,
+                                selectCardCoroutine: SelectCardCoroutine,
+                                afterSelectCardCoroutine: null,
+                                message: $"Select 1 [Seven Code] trait Digimon card from your link cards to place as material ({remaining} remaining).",
+                                maxCount: 1,
```
hooperk: After selecting a digimon, allow multi-select. A valid play is stripping 6 cards off your first Dantemon to play a second

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+                                selectedCard = cardSource;
+                                yield return null;
+                            }
+
+                            yield return ContinuousController.instance.StartCoroutine(selectCardEffect.Activate());
+
+                            if (selectedCard == null) continue;
+
+                            materialCards.Add(selectedCard);
+                            remaining--;
```
hooperk: AfterSelectCardCoroutine can do remaining -= cardSources.Count one multi-select is enabled

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
```diff
@@ -0,0 +1,317 @@
...
+                    {
+                        yield return ContinuousController.instance.StartCoroutine(selectedTarget.AddDigivolutionCardsBottom(materialCards, activateClass));
+                    
+                        yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DigivolveIntoHandOrTrashCard(
+                            selectedTarget,
+                            CanSelectDantemonCondition,
+                            payCost: false,
+                            reduceCostTuple: null,
+                            fixedCostTuple: null,
+                            ignoreDigivolutionRequirementFixedCost: 1,
```
hooperk:
```suggestion
                            ignoreDigivolutionRequirementFixedCost: -1,
```
Leaving as positive 1 will make it cost the player 1 to ignore the requirements

hooperk (changes requested):

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Done — moved the materialCards check into CanSelectOtherBattleAreaCondition itself, matching CanSelectLinkedOrTrashCondition's shape.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Done — maxCount is now Math.Min(remaining, MatchConditionPermanentCount(...)) with canEndNotMax: true.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Done — restructured into 2 steps: select 1 Digimon (matching CanSelectLinkPermanentCondition), then multi-select from that Digimon's own LinkedCards specifically, instead of pooling every Digimon's linked cards into one flat list.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Done — same Math.Min(remaining, ...) + canEndNotMax: true treatment for the trash branch.

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Done — after selecting the Digimon, the link-card selection itself now allows multi-select up to Math.Min(remaining, linkPool.Count).

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: Done — link and trash selections now use afterSelectCardCoroutine to get the full batch and do remaining -= cardSources.Count, instead of the old single-card selectCardCoroutine + manual decrement. Also switched canTargetCondition_ByPreSelecetedList to check the in-progress preSelected list directly (needed now that multi-select is enabled, so the same card can't be picked twice within one batch before materialCards catches up).

Assets/Scripts/CardEffect/BT26/White/BT26_102.cs
alyscs: This was already -1 on the current branch — looks like it got fixed in an earlier round before this review pass reached it. Confirmed still -1 after this push.

PR #1819, opened 2026-08-20
Add BT26-081 Mervamon card effect

alyscs:
## Summary
- Implements BT26-081 Mervamon's card effect (text supplied directly, since the linked GitHub issue's attached image is Cherubimon's art, not Mervamon's).
- Alternate digivolution requirement: Lv.5 w/[TS] trait, cost 4.
- [On Play] [When Digivolving] You may play up to 8 play cost's total worth of [Iliad] trait cards from your hand or trash without paying the cost. Then, to 1 of your opponent's Digimon, give -4000 DP until their turn ends for each of your [Iliad] or [TS] trait Digimon or Tamers.
- [All Turns] All of your [Iliad] trait Digimon gain Alliance, Reboot, Blocker and +2000 DP.
- Assembly -5 (Digimon card with [Minervamon] in its name).

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Purple/BT26_081.cs
+++ b/Assets/Scripts/CardEffect/BT26/Purple/BT26_081.cs
@@ -53,7 +53,7 @@ namespace DCGO.CardEffects.BT26
             {
                 bool CanSelectCardCondition(CardSource cardSource)
                     => cardSource.EqualsTraits("Iliad")
-                        && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass, root: root, isPlayOption: true);
+                        && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass, root: root);
 
                 if (!CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CanSelectCardCondition) && root == SelectCardEffect.Root.Trash) yield break;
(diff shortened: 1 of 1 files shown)


x89rt2:
@hooperk need a check on the play up to 8 cost

alyscs:
Investigated the flagged 'play up to 8 cost' step before review. Confirmed a real gap: [Iliad] Option cards actually exist (BT24-090/094 "Throne Room", Type: [Iliad, TS]), and the code wasn't handling them — CanPlayAsNewPermanent was missing isPlayOption: true (silently excluding Option candidates), and even once selectable, the play step only called PlayPermanentCards on the whole batch, which doesn't support Options. Split the play step into permanentCards/optionCards, routed through PlayPermanentCards/PlayOptionCards respectively.

x89rt2:
> Investigated the flagged 'play up to 8 cost' step before review. Confirmed a real gap: [Iliad] Option cards actually exist (BT24-090/094 "Throne Room", Type: [Iliad, TS]), and the code wasn't handling them — CanPlayAsNewPermanent was missing isPlayOption: true (silently excluding Option candidates), and even once selectable, the play step only called PlayPermanentCards on the whole batch, which doesn't support Options. Split the play step into permanentCards/optionCards, routed through PlayPermanentCards/PlayOptionCards respectively.

Options can never be played. Unlearn the isPlayOption: true thing. That's only used on option cards that "place" themselves in the battle area and the current naming scheme for it is not proper. Revert the change and leave as was before until hooperk says otherwise.

PR #1821, opened 2026-08-20
Add BT26-055 Giromon card effect

alyscs:
## Summary
- Implements BT26-055 Giromon's card effect (text supplied directly; the repo's issue for this card still shows "Not revealed").
- Alternate digivolution requirement: Lv.4 w/[DM] trait, cost 3.
- Fragment <2>.
- [On Play] [When Digivolving] [Counter] [Once Per Turn] You may place 1 card in your hand face down as this Digimon's bottom digivolution card. Then, you may delete 1 of your Digimon with the [Ver.3] trait and all of your opponent's Digimon with the lowest play cost.
- Inherited [All Turns] [Once Per Turn] When this Digimon would leave the battle area, trash your opponent's top security card.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Black/BT26_055.cs
+++ b/Assets/Scripts/CardEffect/BT26/Black/BT26_055.cs
@@ -1,0 +1,192 @@
+using System.Collections;
+using System.Collections.Generic;
+
+// Giromon
+namespace DCGO.CardEffects.BT26
+{
+    public class BT26_055 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Alternate Digivolution Requirement
+            if (timing == EffectTiming.None)
+            {
+                static bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.HasLevel && targetPermanent.TopCard.Level == 4 && targetPermanent.TopCard.HasDMTraits;
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 3, ignoreDigivolutionRequirement: false, card: card, condition: null));
+            }
+            #endregion
+
+            #region Fragment
+            if (timing == EffectTiming.WhenPermanentWouldBeDeleted)
+            {
+                string EffectDescription()
+                    => "<Fragment <2>> (When this Digimon would be deleted, by trashing any 2 of its digivolution cards, it isn't deleted.)";
+
+                cardEffects.Add(CardEffectFactory.FragmentSelfEffect(isInheritedEffect: false, card: card, condition: null, trashValue: 2, effectName: "Fragment <2>", effectDiscription: EffectDescription()));
+            }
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Black/BT26_055.cs
```diff
@@ -0,0 +1,192 @@
...
+                    };
+
+                    GManager.instance.userSelectionManager.SetBoolSelection(selectionElements: selectionElements, selectPlayer: card.Owner, selectPlayerMessage: "Delete 1 of your [Ver.3] trait Digimon and all of your opponent's Digimon with the lowest play cost?", notSelectPlayerMessage: "The opponent is choosing whether to delete Digimon.");
+
+                    yield return ContinuousController.instance.StartCoroutine(GManager.instance.userSelectionManager.WaitForEndSelect());
+
+                    if (GManager.instance.userSelectionManager.SelectedBoolValue)
+                    {
+                        if (hasOwnVer3)
+                        {
```
x89rt2, line 110: This selection doesn't need to exists, you can just select no on the select your own digimon below.

Assets/Scripts/CardEffect/BT26/Black/BT26_055.cs
```diff
@@ -0,0 +1,192 @@
...
+
+                            selectVer3Effect.SetUpCustomMessage("Select 1 [Ver.3] trait Digimon to delete.", "The opponent is selecting 1 [Ver.3] trait Digimon to delete.");
+
+                            yield return ContinuousController.instance.StartCoroutine(selectVer3Effect.Activate());
+                        }
+
+                        if (CardEffectCommons.HasMatchConditionPermanent(IsOpponentDigimon))
+                        {
+                            bool IsMinCostOpponentDigimon(Permanent permanent)
+                                => IsOpponentDigimon(permanent) && CardEffectCommons.IsMinCost(permanent, card.Owner.Enemy, true);
```
x89rt2: This ismincostopponentdigimon could just be what you have in lines 46 and 47.

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/Black/BT26_055.cs
alyscs: Fixed — removed the extra selection; now using `canNoSelect: true` directly on the [Ver.3] select-to-destroy step, so declining there is the only "no" needed.

Assets/Scripts/CardEffect/BT26/Black/BT26_055.cs
alyscs: Fixed — `IsMinCostOpponentDigimon` now reuses `IsOpponentDigimon` (the same condition from lines 46-47) instead of re-deriving it.

x89rt2: I meant the reverse of this. Have the ismincostopponentdigimon be in lines 46-47

alyscs:
Fixed the right way this time — folded the min-cost check directly into IsOpponentDigimon (lines 46-47), and the deletion filter now reuses that same function instead of a separate IsMinCostOpponentDigimon. Doesn't change hasOpponentDigimon's existence-check behavior since any nonempty set always has at least one lowest-cost member.

x89rt2 (approved):

PR #1822, opened 2026-08-20
Add BT26-060 Chronomon: Destroy Mode card effect

### LINKED ISSUE #1643: BT26-060: Chronomon: Destroy Mode
Image

alyscs:
## Summary
- Implements BT26-060 Chronomon: Destroy Mode's card effect (text supplied directly; the repo's issue for this card still shows "Not revealed").
- Stacked on top of #1732 (Yokomon), since this card's "[All Turns] When your effects add to decks" ability needs the `OnAddLibraryAnyone` engine timing introduced there — won't show a clean diff until that PR merges.
- Alternate digivolution requirement: Lv.6 w/[Chronomon] in text or [Giant Slayer] in text, cost 5.
- Security A. +1. Reboot. Blocker.
- Succession (Lv.6 Digimon card with [Chronomon] in its name).
- [On Play] [When Digivolving] Return the top 5 stacked cards of 3 of your opponent's Digimon to the top of the deck.
- [All Turns] [Once Per Turn] When your effects add to decks, you may delete 1 of your opponent's Digimon.


Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,204 @@
...
+            #region Alternate Digivolution Requirement
+            if (timing == EffectTiming.None)
+            {
+                static bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return (targetPermanent.TopCard.HasLevel && targetPermanent.TopCard.Level == 6 && targetPermanent.TopCard.HasText("Chronomon"))
+                        || targetPermanent.TopCard.ContainsCardName("Giant Slayer");
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(permanentCondition: PermanentCondition, digivolutionCost: 5, ignoreDigivolutionRequirement: false, card: card, condition: null));
```
hooperk, line 34: Level should be supplied here, not in permanent condition

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,204 @@
...
+
+                        int returnCount = Math.Min(5, selectedPermanent.DigivolutionCards.Count);
+
+                        if (returnCount >= 1)
+                        {
+                            List<CardSource> returnedCards = selectedPermanent.DigivolutionCards.GetRange(0, returnCount);
+                            returnedCards.Reverse();
+
+                            yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddLibraryTopCards(returnedCards));
+                        }
```
hooperk: This section needs new core code, working like ITrashStack.TrashStack() and checking against a new immunity which Aegiochusmon: Holy Mode will have

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,204 @@
...
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect(SharedEffectName(), CanUseCondition, card);
+                activateClass.SetUpActivateClass((hash) => SharedCanActivateCondition(hash, activateClass), (hash) => SharedActivateCoroutine(hash, activateClass), -1, false, SharedEffectDescription("When Digivolving"));
+                cardEffects.Add(activateClass);
+
+                bool CanUseCondition(Hashtable hashtable)
+                    => CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass)
+                        && CardEffectCommons.CanTriggerWhenDigivolving(hashtable, card);
+            }
+            #endregion
```
hooperk, line 131: Make use of CardEffectFactory.ActivateClassesForSharedEffects as seen on recent cards

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,204 @@
...
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("May delete 1 opponent's Digimon", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDescription());
+                cardEffects.Add(activateClass);
+
+                string EffectDescription()
+                    => "[All Turns] [Once Per Turn] When your effects add to decks, you may delete 1 of your opponent's Digimon.";
+
+                bool CardSourceCondition(CardSource cardSource) => cardSource.Owner == card.Owner;
```
hooperk: Should trigger when adding to either deck, this only triggers for adding own cards to deck

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,204 @@
...
+                        canEndNotMax: false,
+                        selectPermanentCoroutine: null,
+                        afterSelectPermanentCoroutine: null,
+                        mode: SelectPermanentEffect.Mode.Destroy,
+                        cardEffect: activateClass);
+
+                    selectDeleteEffect.SetUpCustomMessage("Select 1 Digimon to delete.", "The opponent is selecting 1 Digimon to delete.");
+
+                    yield return ContinuousController.instance.StartCoroutine(selectDeleteEffect.Activate());
+                }
```
hooperk, line 189: If they choose not to delete, should mark the once per turn as not used with activateClass.RemoveUse()

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,204 @@
...
+                        && CardEffectCommons.CanTriggerWhenDigivolving(hashtable, card);
+            }
+            #endregion
+
+            #region All Turns - Reactive Delete
+            if (timing == EffectTiming.OnAddLibraryAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("May delete 1 opponent's Digimon", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDescription());
```
hooperk: Since first user decision allows no select, we prefer to set optional: false here and activateClass.SetIsSkippable(true); This reduces the number of clicks the user has to make, while still allowing the option to dismiss in bulk

hooperk (changes requested):

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
hooperk: Giant slayer should then be a second add digivolution condition since it does not have level

alyscs: Fixed — split into two separate AddSelfDigivolutionRequirementStaticEffect calls: level: 6 is now passed as its own parameter for the [Chronomon] text branch, and Giant Slayer is a second, level-less alt-digivolve condition (per your follow-up comment below).

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
alyscs: Added IReturnStackToLibrary to CardController.cs, mirroring ITrashStack's shape. It reuses ImmuneFromStackTrashing as the closest existing immunity for now — Aegiochusmon: Holy Mode isn't written yet, so I held off on a speculative dedicated immunity flag; happy to revisit once that card exists.

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
alyscs: Done — On Play/When Digivolving now goes through CardEffectFactory.ActivateClassesForSharedEffects.

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
alyscs: Fixed — passing null as the cardSourceCondition to CanTriggerOnAddLibrary so it triggers when either player's deck is added to.

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
alyscs: Added isUsed tracking + activateClass.RemoveUse() for the no-selection path.

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
alyscs: Switched to optional: false + activateClass.SetIsSkippable(true).

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
hooperk: I believe it would be easiest to create the effect as part of this branch, so the IReturnStackToLibrary can reference it, and then it is little effort to simply use the new effect on Aegiochusmon: Holy Mode, rather than needing to return to rework this card's changes later

alyscs: Done — added a dedicated IImmuneFromStackReturnToLibraryEffect interface, ImmuneStackReturnToLibraryClass, CardEffectFactory.CanNotBeReturnedToLibraryBySkillStaticEffect, and Permanent.ImmuneFromStackReturnToLibrary, mirroring the existing stack-trashing immunity infra. IReturnStackToLibrary now checks this dedicated flag instead of reusing ImmuneFromStackTrashing. Aegiochusmon: Holy Mode can use CardEffectFactory.CanNotBeReturnedToLibraryBySkillStaticEffect directly whenever it's written.

alyscs:
Following up on the AddLibraryAnyone discussion from #1732 (Yokomon): this card had the same class of bug, actually worse — CanUseCondition passed null for cardCondition with no ownership check at all, so it triggered on ANY card added to ANY deck by anyone, not just the owner's effects. Fixed it to use the same CardEffectCommons.IsOwnerEffect(cardEffect, card) check now that AddLibraryTopCards/BottomCards thread an ICardEffect through the hashtable.

Assets/Scripts/Script/CardController.cs
```diff
@@ -5905,6 +5905,56 @@ public IEnumerator Overflow()
...
 
+#region ReturnStackToLibrary
+
+public class IReturnStackToLibrary
+{
+    /// <summary>
+    /// Return Stack To Library Class
+    /// </summary>
+    /// <param name="permanent">Target Permanent</param>
+    /// <param name="ReturnCount">Amount of digivolution cards to return to the top of the library</param>
```
hooperk:
```suggestion
    /// <param name="ReturnCount">Amount of stacked cards to return to the top of the library</param>
```

Assets/Scripts/Script/CardController.cs
```diff
@@ -5905,6 +5905,56 @@ public IEnumerator Overflow()
...
+        if (_permanent.ImmuneFromStackReturnToLibrary(_cardEffect)) yield break;
+        if (_permanent.TopCard.CanNotBeAffected(_cardEffect)) yield break;
+
+        int count = Math.Min(_permanent.DigivolutionCards.Count, _returnCount);
+
+        if (count >= 1)
+        {
+            List<CardSource> returnedCards = _fromTop
+                ? _permanent.DigivolutionCards.GetRange(0, count)
+                : _permanent.DigivolutionCards.GetRange(_permanent.DigivolutionCards.Count - count, count);
```
hooperk:
```suggestion
            List<CardSource> returnedCards = _permanent.StackCards.GetRange(0, count);
```
Bounces Stacked cards (topmost first) not digivolution cards

Assets/Scripts/Script/CardController.cs
```diff
@@ -5905,6 +5905,56 @@ public IEnumerator Overflow()
...
+    /// <param name="permanent">Target Permanent</param>
+    /// <param name="ReturnCount">Amount of digivolution cards to return to the top of the library</param>
+    /// <param name="cardEffect">Card Effect</param>
+    /// <param name="fromTop">Should the returned cards start from the top of the stack, true by default</param>
+    public IReturnStackToLibrary(Permanent permanent, int ReturnCount, ICardEffect cardEffect, bool fromTop = true)
+    {
+        _permanent = permanent;
+        _returnCount = ReturnCount;
+        _cardEffect = cardEffect;
+        _fromTop = fromTop;
```
hooperk: Effect can only ever be from top

hooperk (changes requested):

alyscs:
Rebased against develop now that Yokomon (#1732) is merged, and addressed the latest CardController.cs comments: IReturnStackToLibrary now always works from the top (removed the dead fromTop parameter) and bounces StackCards (topmost first, including the top card itself) instead of only DigivolutionCards underneath it. Also threads cardEffect through to AddLibraryTopCards now that it's a required parameter.

Assets/Scripts/Script/CardController.cs
alyscs: Applied — doc comment updated.

Assets/Scripts/Script/CardController.cs
alyscs: Fixed — now reads from StackCards instead of DigivolutionCards, so the top card itself is included, not just the material underneath it.

Assets/Scripts/Script/CardController.cs
alyscs: Applied — removed the fromTop parameter entirely and switched to StackCards.GetRange(0, count), matching your suggestion.

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -0,0 +1,196 @@
...
+            {
+                cardEffects.Add(CardEffectFactory.BlockerSelfStaticEffect(isInheritedEffect: false, card: card, condition: null));
+            }
+            #endregion
+
+            #region Succession
+            if (timing == EffectTiming.None)
+            {
+                bool CardCondition(CardSource cardSource)
+                    => cardSource.HasLevel && cardSource.Level == 6 && cardSource.ContainsCardName("Chronomon");
```
hooperk, line 63: Minor FYI: CardSource.IsLevel6 is cleaner for exact level, it does the HasLevel inside of it. WE really only need to check HasLevel and use Level directly when doing greater than/less than comparisons

hooperk (approved):

PR #1827, opened 2026-08-21
Same copy effect different top card patch

sollybr:
Attempt of dealing with the 1-1 relationship between the 'controller of card effects' (`UseEffectsThisTurn` owner) and the 'card entity'

---

**TODO**:
- Evaluate possible duplication of entries on digivolution
- Verify complexity of replacing `UseEffectsThisTurn` with a proper dictionary in `TurnStateMachine` that maps permanents to its used effects in X turn

Continuity to: #1729


sollybr:
@hooperk need a review on this one so hashstrings can do their job identifying used effects through further evolutions

PR #1828, opened 2026-08-21
Offline Bot Matches

sollybr:
"this should make certain approaches to debugging easier; connection issues post bot match 'start' for simple tests should not stop the match progress any longer" @ discord.


Assets/Scripts/Script/SelectBattleMode.cs
```diff
@@ -134,6 +134,16 @@ void StartSelectBattleDeck(bool isAI)
         ContinuousController.instance.isAI = isAI;
         ContinuousController.instance.isRandomMatch = true;
 
+        if (!isAI && PhotonNetwork.OfflineMode)
```
hooperk, line 137: I don't see where OfflineMode is being set to true in the first place?

sollybr:
Continuing on #1833 with old history

PR #1834, opened 2026-08-22
Same copy effect different top card patch

sollybr:
Workaround for copied used effects list being bound to effect controller of the top most card (source) in a permanent
---
Attempt of dealing with the 1-1 relationship between the 'controller of card effects' (`UseEffectsThisTurn` owner) and the 'card entity'

---

**TODO**:
- Evaluate possible duplication of entries on digivolution
- Verify complexity of replacing `UseEffectsThisTurn` with a proper dictionary in `TurnStateMachine` that maps permanents to its used effects in X turn

Continuity to: #1827


Assets/Scripts/Script/CardController.cs
```diff
@@ -1391,6 +1391,8 @@ public IEnumerator PlayPermanent()
 
                             PlayLog.OnAddLog?.Invoke($"\nEvolution:\n{card.BaseENGCardNameFromEntity}({card.CardID})\n");
 
+                            card.cEntity_EffectController.UseEffectsThisTurn.AddRange(_targetPermanent.TopCard.cEntity_EffectController.UseEffectsThisTurn);
```
hooperk, line 1394: This will mess with a level 5 usign the once per turn of a level 4 beneath it, from which it is copying

hooperk: I am thinking instead the issue is that our CopiedEffects are adding to the top card at all times... instead they should add to the source card (as inherited/Link effects in those cases)

hooperk: Pushing my own thought on this, I think if we just ensure we copy the effects to the card with the copy effect rather than the top card everything will behave correctly

x89rt2:
@sollybr @hooperk https://github.com/DCGO2/DCGO/pull/1838 Did not seem to fix the issue according to doc.

Actually according the messages, it broke: https://discord.com/channels/1095717982522585098/1392137159275708456/1541610987927507135

Or might be unrelated? I reverted and it does not use the OP currently in develop build

sollybr:
> This will mess with a level 5 usign the once per turn of a level 4 beneath it, from which it is copying

You have card examples so we can do some tests? @hooperk

x89rt2:
> > This will mess with a level 5 usign the once per turn of a level 4 beneath it, from which it is copying
> 
> You have card examples so we can do some tests? @hooperk

? You want an example of a set up that fails with this? Get have a stack of ex12 betel, a lvl 5 that has an inherited to copy effects, and a lvl 6. Use the main effect of betel while you are level 6, then evo into level 7. You should not be able to use the Main effect of betel at that point, but current code allows you to.

x89rt2:
whoops

sollybr:
> You want an example of a set up that fails with this? Get have a stack of ex12 betel, a lvl 5 that has an inherited to copy effects, and a lvl 6. Use the main effect of betel while you are level 6, then evo into level 7. You should not be able to use the Main effect of betel at that point, but current code allows you to.

That was this PR's motivation.

Assets/Scripts/Script/CardController.cs
hooperk: actually it would be fine, we give it a different hashstring for our copied effects

hooperk (approved):

Assets/Scripts/Script/CardController.cs
hooperk: Wait, this will cause problems for some non-copy effects

hooperk (changes requested):

Assets/Scripts/Script/CardController.cs
hooperk: Imagine Sakuyamon deck, They digivolve into ST Sakuyamon, use the OPT When Digivolving/When attacking. They then digivolve into Maid mode and digivolve maid mode into sakuyamon again.  Because of this line, this new Sakuyamon has an entry int its used effects for the WD/WA, and so it cannot activate again

sollybr:
> Wait, this will cause problems for some non-copy effects

That makes more sense; you have some to test? LF/NC maybe?
We can probably do a proper filter in the range of additions so it doesn't affect others.

Assets/Scripts/Script/CardController.cs
sollybr:
```suggestion
foreach (ICardEffect effect in _targetPermanent.TopCard.cEntity_EffectController.UseEffectsThisTurn)
{
    if (effect.OriginalEffectSourceCard is not null)
    {
        card.cEntity_EffectController.UseEffectsThisTurn.Add(effect);
    }
}
```

sollybr: > Imagine Sakuyamon deck, They digivolve into ST Sakuyamon, use the OPT When Digivolving/When attacking. They then digivolve into Maid mode and digivolve maid mode into sakuyamon again. Because of this line, this new Sakuyamon has an entry int its used effects for the WD/WA, and so it cannot activate again

Tried the filter approach on latest review

sollybr: @hooperk should @x89rt2 apply this?

sollybr (commented):
xeight's update ought to limit this patch to copied effects which have unique hashstring identifiers

Issue #1864, opened 2026-08-27
EX13-020: Magnamon


hooperk:
<img width="881" height="1200" alt="Image" src="https://github.com/user-attachments/assets/0cbb18c5-8bba-41dd-8f5b-113d84d8afcb" />

PR #1927, opened 2026-08-27
Add EX13-020 Magnamon card effect

alyscs:
## Summary
First card of the EX13 set (issue #1864). Creates the `Assets/Scripts/CardEffect/EX13/` folder and adds:

- Alternate digivolve requirement: any card named [Veemon], cost 3
- Blocker, Armor Purge
- Assembly -2 [Veemon]
- [On Play] [When Digivolving] [When Attacking] [Once Per Turn]: this Digimon gets +1000 DP until your opponent's turn ends for each distinct color across both trashes. Then, to 1 of your opponent's Digimon, give -4000 DP until their turn ends for every 5000 DP this Digimon has.
- [End of Your Turn] [Once Per Turn] (also Inherited): 1 of your Digimon with the [Free] or [Royal Knight] trait may unsuspend.

No CardBaseEntity data included, matching current convention (created in bulk closer to beta).

### DIFF
--- a/Assets/Scripts/CardEffect/EX13/Blue/EX13_020.cs
+++ b/Assets/Scripts/CardEffect/EX13/Blue/EX13_020.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
@@ -103,8 +102,6 @@ namespace DCGO.CardEffects.EX13
 
                 if (CardEffectCommons.HasMatchConditionPermanent(CanSelectDebuffTargetCondition))
                 {
-                    int maxCount = Math.Min(1, CardEffectCommons.MatchConditionPermanentCount(CanSelectDebuffTargetCondition));
-
                     SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
 
                     selectPermanentEffect.SetUp(
@@ -112,7 +109,7 @@ namespace DCGO.CardEffects.EX13
                         canTargetCondition: CanSelectDebuffTargetCondition,
                         canTargetCondition_ByPreSelecetedList: null,
                         canEndSelectCondition: null,
-                        maxCount: maxCount,
+                        maxCount: 1,
                         canNoSelect: false,
                         canEndNotMax: false,
                         selectPermanentCoroutine: SelectPermanentCoroutine,
@@ -162,7 +159,8 @@ namespace DCGO.CardEffects.EX13
 
                 bool CanSelectPermanentCondition(Permanent permanent)
                     => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
-                        && (permanent.TopCard.ContainsTraits("Free") || permanent.TopCard.ContainsTraits("Royal Knight"));
+                        && (permanent.TopCard.ContainsTraits("Free") || permanent.TopCard.ContainsTraits("Royal Knight"))
+                        && CardEffectCommons.CanUnsuspend(permanent);
 
                 bool CanUseCondition(Hashtable hashtable)
                     => CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass)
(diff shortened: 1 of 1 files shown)


alyscs:
Self-review note: "for each color in trashes" has no existing precedent in the codebase (no other card phrases it this way, plural with no possessive). Implemented as distinct colors across both players' trashes combined, since that's the most natural reading of the plural with no "your"/"their" qualifier — but this is an interpretation, not a verified ruling. Worth double-checking against official text once it's confirmed, in case it should be scoped to just your own trash.

Assets/Scripts/CardEffect/EX13/Blue/EX13_020.cs
```diff
@@ -0,0 +1,202 @@
...
+
+            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
+            {
+                Permanent thisPermanent = card.PermanentOfThisCard();
+
+                yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.ChangeDigimonDP(targetPermanent: thisPermanent, changeValue: 1000 * TrashColorCount(), effectDuration: EffectDuration.UntilOpponentTurnEnd, activateClass: activateClass));
+
+                if (CardEffectCommons.HasMatchConditionPermanent(CanSelectDebuffTargetCondition))
+                {
+                    int maxCount = Math.Min(1, CardEffectCommons.MatchConditionPermanentCount(CanSelectDebuffTargetCondition));
```
x89rt2: does not need math.min, there will always only bee 1 target at this stage since it already has a match condition if above

Assets/Scripts/CardEffect/EX13/Blue/EX13_020.cs
```diff
@@ -0,0 +1,202 @@
...
+                activateClass.SetIsInheritedEffect(true);
+                activateClass.SetHashString("EX13_020_EndOfTurn");
+                cardEffects.Add(activateClass);
+
+                string EffectDescription()
+                    => "[End of Your Turn] [Once Per Turn] 1 of your Digimon with the [Free] or [Royal Knight] trait may unsuspend.";
+
+                bool CanSelectPermanentCondition(Permanent permanent)
+                    => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
+                        && (permanent.TopCard.ContainsTraits("Free") || permanent.TopCard.ContainsTraits("Royal Knight"));
```
x89rt2: Since this is an optional effect, it's best to make this one as automatic as possible so do !permanent.issuspended && permanent.canunsuspend. For mandatory effects, we would not want to do this so we can choose to whiff the effects on purpose.

Assets/Scripts/CardEffect/EX13/Blue/EX13_020.cs
```diff
@@ -0,0 +1,202 @@
...
+                        selectPermanentCoroutine: null,
+                        afterSelectPermanentCoroutine: null,
+                        mode: SelectPermanentEffect.Mode.UnTap,
+                        cardEffect: activateClass);
+
+                    selectPermanentEffect.SetUpCustomMessage("Select 1 [Free]/[Royal Knight] Digimon to unsuspend.", "The opponent is selecting 1 [Free]/[Royal Knight] Digimon to unsuspend.");
+
+                    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+                }
+            }
```
x89rt2, line 204: Since it's OPT and optional, to account for every possibility, let's do the isUsed = false; stuff we do in some other OPT effects for this one.

x89rt2 (changes requested):

alyscs:
Addressed all 3 points:
1. Dropped the Math.Min — maxCount is now just 1 directly, since the surrounding HasMatchConditionPermanent already guarantees a match.
2. The End of Your Turn unsuspend now filters to CardEffectCommons.CanUnsuspend(permanent) (actually suspended + able to unsuspend), matching the optional-vs-mandatory targeting distinction.
3. Added isUsed/RemoveUse() tracking so declining or having no valid pick doesn't burn the Once Per Turn use.

x89rt2 (approved):

PR #1943, opened 2026-08-29
Fix BT26-103 (Jupitermon: Wrath Mode) Succession never triggering

alyscs:
## Summary
- Wrath Mode's Succession self-check used `EqualsCardName("Jupitermon")`, but this card's own printed name is `"Jupitermon: Wrath Mode"` (same colon-joined display-name convention as other dual/mode cards), so the exact-match check could never succeed and Succession never activated.
- Switched to `ContainsCardName("Jupitermon")`, matching the same pattern already used by other Succession cards (e.g. BT26_060 checks `ContainsCardName("Chronomon")` against its own name `"Chronomon: Destroy Mode"`).

## Test plan
- [ ] Digivolve Wrath Mode onto a Digimon named "Jupitermon" and confirm Succession grants that card's effects.


x89rt2:
This is just incorrect.

PR #1945, opened 2026-08-29
BT26-060 (Chronomon: Destroy Mode): return-order picker + return-count fixes

alyscs:
## Summary
- The "Return the top 5 stacked cards" ability returned cards without letting the player choose the order the opponent will draw them back in. Added an order-choice selection so the player controls which card the opponent draws first vs. second, etc.
- Fixed an `InvalidOperationException` crash (`Permanent.ImmuneFromStackReturnToLibrary`'s cross-board scan) that could occur mid-loop if a previously-processed permanent's board state changed the availability of a later one — added a defensive existence re-check per permanent in the return-order loop.
- Fixed the return-count/slice logic: it originally computed the return count from `Permanent.StackCards` (which includes the target's own active top card, not just digivolution material), so a Digimon with zero digivolution cards could have its own card returned to the deck, wiping it off the field entirely. Fixed to return from the top of the stack (top card included, matching "top 5 stacked cards") while always leaving at least 1 card behind so the permanent is never fully removed, and excluded zero-material Digimon from being selectable as a target at all (since returning from them would always be a no-op).

## Test plan
- [ ] Activate the ability against an opponent's Digimon with 3+ stacked cards and confirm the order-choice prompt lets you pick which cards return, and in what order.
- [ ] Confirm a target with zero digivolution cards is not offered as a selectable target.
- [ ] Confirm returning from a 3-card stack (e.g. top + 2 material) can return the top 2 cards, leaving 1 behind, without crashing.

### DIFF
--- a/Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
+++ b/Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
@@ -112,6 +112,7 @@ namespace DCGO.CardEffects.BT26
                     foreach (Permanent selectedPermanent in selectedPermanents)
                     {
                         if (selectedPermanent == null) continue;
+                        if (!CardEffectCommons.IsPermanentExistsOnBattleArea(selectedPermanent)) continue;
                         if (selectedPermanent.TopCard == null) continue;
                         if (selectedPermanent.ImmuneFromStackReturnToLibrary(activateClass)) continue;
                         if (selectedPermanent.TopCard.CanNotBeAffected(activateClass)) continue;
--- a/Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
+++ b/Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
@@ -74,8 +74,7 @@ namespace DCGO.CardEffects.BT26
                 => $"[{tag}] Return the top 5 stacked cards of 3 of your opponent's Digimon to the top of the deck.";
 
             bool CanSelectTargetCondition(Permanent permanent)
-                => CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card)
-                    && !permanent.HasNoDigivolutionCards;
+                => CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card);
 
             IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
             {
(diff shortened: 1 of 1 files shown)


Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -74,7 +74,8 @@ string SharedEffectDescription(string tag)
                 => $"[{tag}] Return the top 5 stacked cards of 3 of your opponent's Digimon to the top of the deck.";
 
             bool CanSelectTargetCondition(Permanent permanent)
-                => CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card);
+                => CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card)
+                    && !permanent.HasNoDigivolutionCards;
```
x89rt2: Incorrect. You are legally allowed to select a digimon with no digivolution sources. It would just not do anything, but it's legal.

x89rt2 (changes requested):

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
alyscs: Fixed — removed the `&& !permanent.HasNoDigivolutionCards` filter from `CanSelectTargetCondition`. This ability has no decline point (mandatory, not optional), so the target pool now matches the printed text's full targeting breadth. The existing `returnCount <= 0 -> continue` guard in the resolution loop already handles a zero-material target as a safe no-op.

Assets/Scripts/CardEffect/BT26/Black/BT26_060.cs
```diff
@@ -111,7 +111,61 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
...
+
+                        IEnumerator AfterSelectOrderCoroutine(List<CardSource> cardSources)
+                        {
+                            drawOrderedCards = cardSources.Clone();
+                            yield return null;
+                        }
+
+                        drawOrderedCards.Reverse();
+
+                        yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddLibraryTopCards(drawOrderedCards, cardEffect: activateClass));
```
hooperk: This changes should be happening in IReturnStackToLibrary, not the card itself

x89rt2 (approved):

PR #1946, opened 2026-08-29
Fix Succession keyword: OPT tracking, self-poisoning identity, and parameter mismapping

alyscs:
## Summary
Three real bugs found while testing the [Succession] keyword (Chronomon: Destroy Mode digivolving onto Chronomon: Holy Mode, same-turn), all in shared `CopyDigivolutionCardEffects`/`AddSkillClass` infrastructure used by every Succession card (and by 9 other cards that call `CopyDigivolutionCardEffects` directly):

1. **Once-per-turn tracking was shared with the source card.** `CopyDigivolutionCardEffects` reused the source card's own `ActivateClass` object reference (only mutating its conditions/hash in place), and `ICardEffect.IsSameEffect` short-circuits on reference equality — so a `[Once Per Turn]` effect used on the source card earlier the same turn silently couldn't trigger again when copied onto a new card via Succession, even though it should be an independently-tracked copy. Fixed by building a genuinely new `ActivateClass` with its own `EffectSourceCard`/hash, delegating activation behavior to the original.
2. **`AddSkillClass.GetCardEffect` self-poisoning.** The engine's cross-permanent scan probes every `AddSkillClass` wrapper on the field against whatever card is currently being evaluated for an event — most such probes are expected to fail the wrapper's own condition. `GetCardEffect` called `SetEffectSourceCard(card)` unconditionally regardless of that result, so a single failed probe against an unrelated card permanently overwrote the wrapper's real identity (self-reinforcing, since later calls read that same field back to decide what to probe next). Fixed to only adopt the probed card's identity when the wrapper actually matched it.
3. **Succession parameter mismapping.** `SuccessionSelfEffect` passed its own `cardCondition` parameter (the printed keyword's material filter, e.g. "Lv.6 with Chronomon in name") into `CopyDigivolutionCardEffects`'s `cardSourceCondition` slot instead of its `cardCondition` slot — silently replacing the real self-check ("am I still my own permanent's top card") with the material filter, and leaving the actual buried-card filter unapplied. Fixed the parameter mapping.

## Test plan
- [ ] Use a Once-Per-Turn When-Digivolving effect on a card, then digivolve a Succession card matching its name onto it the same turn, and confirm the Succession copy can also trigger.
- [ ] Confirm Succession only copies from the correct (topmost matching) buried card, not an unrelated one.

### DIFF
--- a/Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
+++ b/Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
@@ -102,7 +102,17 @@ public partial class CardEffectFactory
 
                     if (cardEffect is ActivateClass activateClass)
                     {
-                        getCardEffects.Add(activateClass);
+                        // Build a brand-new ActivateClass rather than mutating/reusing the source
+                        // card's own instance. The source card's copy needs its own independent
+                        // EffectSourceCard/HashString so per-turn-use tracking (ICardEffect.IsSameEffect,
+                        // which short-circuits on reference equality) doesn't treat "the original card
+                        // already used this ability this turn" as also covering "the Digimon that just
+                        // gained this ability via Succession/copy already used it" -- per game rules,
+                        // gaining another card's effects this way grants an independently-tracked copy,
+                        // not a shared use-count with the original (real bug: a [Once Per Turn] When
+                        // Digivolving effect used earlier the same turn on the source card silently
+                        // couldn't trigger again when copied onto the new top card via Succession, even
+                        // though it's a fresh instance from the new card's perspective).
 
                         List<CardSource> ValidCardSources = null;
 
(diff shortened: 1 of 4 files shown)


Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
```diff
@@ -123,24 +134,36 @@ bool ValidCardSourceAtActivate()
...
+                                && (originalActivateCondition is null || originalActivateCondition(hashtable)),
+                            hashtable => activateClass.Activate(hashtable),
+                            activateClass.MaxCountPerTurn,
+                            activateClass.IsOptional,
+                            activateClass.EffectDescription);
+
+                        copiedActivateClass.SetOriginalEffectSourceCard(activateClass.OriginalEffectSourceCard);
+                        copiedActivateClass.SetHashString(GenerateHashString(card, activateClass.OriginalEffectSourceCard, activateClass.HashString, isInheritedEffect, isLinkedEffect));
+                        copiedActivateClass.SetIsInheritedEffect(isInheritedEffect);
+                        copiedActivateClass.SetIsLinkedEffect(isLinkedEffect);
```
hooperk, line 163: If setting these as inherited link effects, we need to ensure they are being added to the card with this effect, not the Topcard, on line 44. If that doesn';t work then we'll want to set these 2 as false if copying onto the top card

hooperk (changes requested):

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
alyscs: Good catch. `DefaultCardSourceCondition` never actually checked `cardSource == card` -- it only checked "is cardSource the top card of a permanent that equals card's own permanent", so an unrelated co-located card could satisfy that just by being the current top card. And the same check unconditionally required `cardSource == permanent.TopCard`, which is permanently impossible once the copy is meant to be inherited/linked (buried/attached cards are by definition never the top card) -- so the 5 real callers passing `isInheritedEffect: true` (BT15_039, RB1_023/015/009/030, BT10_011) could never actually apply their inherited copy once the host was buried.

Fixed both: added the explicit `cardSource == card` identity check, and only require `cardSource == permanent.TopCard` for the non-inherited/non-linked case -- an inherited or linked copy now applies to `card` regardless of whether it's currently on top. Pushed in 4122a3f2d.

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
```diff
@@ -123,24 +139,36 @@ bool ValidCardSourceAtActivate()
...
+                            hashtable => ValidCardSourceAtTrigger()
+                                && (originalUseCondition is null || originalUseCondition(hashtable)),
+                            card);
+
+                        copiedActivateClass.SetUpActivateClass(
                             hashtable => ValidCardSourceAtActivate()
-                            && (originalActivateCondition is null || originalActivateCondition(hashtable))
-                        );
+                                && (originalActivateCondition is null || originalActivateCondition(hashtable)),
+                            hashtable => activateClass.Activate(hashtable),
```
hooperk, line 155: This will still increment the OPT of the source card as that is handled within Activate. We need ot instead access the Coroutine

hooperk (changes requested):

Assets/Scripts/Script/CardEffectFactory/CopiedEffects.cs
alyscs: Confirmed and fixed — found the concrete evidence: BT26_016 Chronomon: Holy Mode has `if (!isUsed) activateClass.RemoveUse();` inside its ActivateCoroutine. That coroutine is reused verbatim by copiedActivateClass's delegated Activate() call, but the closure still self-references the ORIGINAL activateClass instance -- so RemoveUse()/AddUse() (which operate via the calling instance's own EffectSourceCard + object identity for the UseEffectsThisTurn list) touched the source card's own tracking instead of the copy's, even though copiedActivateClass otherwise has fully independent EffectSourceCard/HashString.

Rather than trying to expose/rebind the private raw coroutine delegate (which wouldn't help on its own -- the closure is compiled against the original activateClass variable and can't be rebound at runtime), I added `ICardEffect.UseTrackingRedirectTarget`: RemoveUse()/AddUse() now check it and, when set, operate against the redirect target's EffectSourceCard + identity instead of their own. CopyDigivolutionCardEffects sets activateClass's redirect target to copiedActivateClass for the duration of the delegated Activate() call and clears it right after, so any self-referencing RemoveUse()/AddUse() inside the reused coroutine now correctly affects the copy's own OPT tracking.

Pushed in fdcc036c5.

alyscs:
Self-reviewed the previous commit (fdcc036c5) and found 2 real robustness issues, fixed in af969a97a:

1. No `try/finally` around the redirect set/clear -- an exception in the reused coroutine, or it being stopped externally, would skip cleanup and leave the redirect permanently pointed at a stale copy, corrupting the source card's own later ability activations.
2. `UseTrackingRedirectTarget` was a single mutable field, not per-invocation state -- if the source ability were mid-activation for one copy (e.g. awaiting player input) when a second, unrelated activation for a different copy started, the second would clobber the first's redirect target.

Fixed both: `UseTrackingRedirectTarget` is now backed by a `Stack<ICardEffect>` (Push/Pop instead of Set/Clear), and the delegated `Activate()` call is wrapped in `try/finally` so the pop always runs.

x89rt2 (approved):

PR #1947, opened 2026-08-29
Fix BT26-100 (Dark Field): [Main] ability could play a Digi-Egg

alyscs:
## Summary
- The `[Main]` ability's card-selection filter (`CanSelectCardCondition`) was missing the `cardSource.IsDigimon` check that this card's own `[Security]` ability (`CanPlayCondition`, same file) correctly has — letting a level ≤4 Titan-trait Digi-Egg slip through and get played as a new permanent via an ability meant only for Digimon cards.
- Added the missing `IsDigimon` check, matching the `[Security]` ability's condition.

## Test plan
- [ ] Confirm a Titan-trait Digi-Egg can no longer be selected as the free-play target for the `[Main]` ability.
- [ ] Confirm a legal level ≤4 Titan Digimon card can still be played as before.


x89rt2:
Not needed. The rule is only things with play costs can be played, and that was fixed shortly before you made this PR by Kage. Your change is technically incorrect.

PR #2005, opened 2026-09-15
Fix OnTrashDigivolutionCards and dancing around multiple locations

x89rt2:


Assets/Scripts/CardEffect/BT11/Purple/BT11_081.cs
```diff
@@ -128,11 +128,11 @@ string EffectDiscription()
 
                 bool CanUseCondition(Hashtable hashtable)
                 {
-                    if (CardEffectCommons.IsExistOnBattleArea(card))
+                    if (CardEffectCommons.IsExistOnBattleAreaTrigger(card, activateClass))
```
hooperk, line 131: Not needed for self trash effects, actual check is what is set within CanTriggerOnTrashSelfDigivolutionCard

hooperk (changes requested):

PR #2015, opened 2026-09-18
Apply copied effects directly instead of addskill

hooperk:
IAddSkillClass has effects it is not allowed to copy. (all with timing == None, so static keywords)
Instead, this method now directly copies all valid effects onto the card, so when it is queried for all of its effects, it returns he copied effects included with its own.


x89rt2:
@sollybr Can you give this a look over?

sollybr:
Only issue found was stack overflow from inherited-effect copies by Gammamon's Perfect/Ultimate. Log appended.
Other archetypes concerned are stable.
[lvl6-copied-effects-overflow.log](https://github.com/user-attachments/files/32445030/lvl6-copied-effects-overflow.log)

x89rt2 (changes requested):
@hooperk above comment

